PR #1, opened 2016-11-03
Feature/msb

hbgwhite:
This change set adds code to support Multi-Stage braille test-takers. It provides a way to retrieve the test's MSB property and pass it to the item selection shell for segment processing.

### DIFF
--- a/student/src/main/java/tds/student/performance/services/impl/TestOpportunityServiceImpl.java
+++ b/student/src/main/java/tds/student/performance/services/impl/TestOpportunityServiceImpl.java
@@ -161,10 +161,14 @@ public class TestOpportunityServiceImpl implements TestOpportunityService {
                     testOpportunity.getClientName(),
                     testOpportunity.getTestId());
 
+            Boolean isMsb = configurationDao.isMsb(
+                    testOpportunity.getClientName(),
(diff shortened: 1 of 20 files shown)


student/src/main/java/tds/student/performance/services/impl/TestOpportunityServiceImpl.java
```diff
@@ -161,10 +161,14 @@ public TestConfig startTestOpportunity(OpportunityInstance opportunityInstance,
                     testOpportunity.getClientName(),
                     testOpportunity.getTestId());
 
+            Boolean isMsb = configurationDao.isMsb(
```
millerts: Does this have to be the Object boolean or could it be the primitive?  I understand the current code base likes to use Objects instead of primitives

millerts (approved):
One small question that doesn't change the meat of the review

student/src/main/java/tds/student/performance/services/impl/TestOpportunityServiceImpl.java
hbgwhite: No, it doesn't. I'm converting it to a primitive.

PR #2, opened 2016-11-22
Feature/braille transcript

jtreuting:
This request alters how student handles an Emboss request.  If student has Braille Transcript and there is a transcript version of the braille file for this item, it changes the request label presented to the proctor to include Transcript in the title, and submits a colon separated list of braille files with the first one being the passage (or item) and the second one being the transcript.

### DIFF
--- a/student/src/main/java/tds/student/services/PrintService.java
+++ b/student/src/main/java/tds/student/services/PrintService.java
@@ -93,6 +93,16 @@ public class PrintService
 
   }
 
+  /**
+   * Determines if the list of attachments means that this is a Braille Transcript embossing request
+   *
+   * @param brailleAttachments List of braille attachments for this request
+   * @return
+     */
+  private static boolean isTranscriptRequest(List<ITSAttachment> brailleAttachments) {
+    return brailleAttachments.size() == 2 && brailleAttachments.get(1).getSubType().endsWith("_transcript");
+  }
+
   public boolean printPassage (OpportunityInstance oppInstance, PageGroup pageGroupToPrint, String requestParameters) throws ReturnStatusException {
     return printPassage ("PRINTPASSAGE", oppInstance, pageGroupToPrint, requestParameters);
   }
@@ -194,7 +204,7 @@ public class PrintService
       List<ITSAttachment> brailleAttachments = content.GetBrailleTypeAttachment (accLookup);
 
       // check if any attachments
-      if (brailleAttachments == null || brailleAttachments.size() == 0) {
+      if (brailleAttachments == null || brailleAttachments.isEmpty()) {
         // log attachments are missing
         String testKey = testOpp.getTestKey ();
         String groupID = pageGroupToPrint.getId ();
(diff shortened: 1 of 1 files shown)


student/src/main/java/tds/student/services/PrintService.java
```diff
@@ -162,24 +173,28 @@ public boolean printPassageBraille (String requestType, TestOpportunity testOpp,
...
-      // get attachemnt for the accommodations braille type
-      ITSAttachment brailleAttachment = content
-          .GetBrailleTypeAttachment (accLookup);
+
+      // get attachments for the accommodations braille type
+      List<ITSAttachment> brailleAttachments = content.GetBrailleTypeAttachment (accLookup);
+
       // check if any attachments
-      if (brailleAttachment == null) {
+      if (brailleAttachments == null || brailleAttachments.size() == 0) {
```
millerts: I like to use `brailleAttachments.isEmpty()` instead of the size comparison.  Reads better.

student/src/main/java/tds/student/services/PrintService.java
```diff
@@ -248,16 +278,30 @@ public boolean printItemBraille (TestOpportunity testOpp, ItemResponse responseT
         _tdsLogger.rendererWarn (error, "printItemBraille");
         return false;
       }
+
+      ITSAttachment mainBrailleAttachment = brailleAttachments.get(0);
+
+      // the subType in the Item Metadata attachmentList is the only way to determine if there is a Braille Transcript file
+      //  it will follow the convention of being named with an _transcript suffix according to the specification
+      boolean isTranscript = brailleAttachments.size() == 2 && brailleAttachments.get(1).getSubType().endsWith("_transcript");
```
millerts: Possibly extract this call out to a private static method that could do the check?  So the conditional lives in a single place.

student/src/main/java/tds/student/services/PrintService.java
```diff
@@ -248,16 +278,30 @@ public boolean printItemBraille (TestOpportunity testOpp, ItemResponse responseT
...
-      String requestValue = brailleAttachment.getFile ();
-      String requestParameters = "FileFormat:"
-          + brailleAttachment.getType ().toUpperCase ();
-      String requestDesc = getItemLabel (responseToPrint);
-      requestDesc = String.format ("%1$s (%2$s)", requestDesc,
-          brailleAttachment.getType ());
-      _oppRepository.submitRequest (testOpp.getOppInstance (),
-          responseToPrint.getPage (), responseToPrint.getPosition (),
-          requestType, requestDesc, requestValue.replace ("\\", "\\\\"), requestParameters);
+      String requestValue = mainBrailleAttachment.getFile();
```
millerts, line 295: Actually, is there a way to extract this code block and the isTranscript check into its own method.  This looks almost identical to lines 216 - 225 above.

millerts (changes requested):

student/src/main/java/tds/student/services/PrintService.java
jtreuting: it is almost but slightly different.  i was trying to keep a smaller-ish footprint since this is editing existing code

millerts: 👍  - Figured

PR #5, opened 2016-12-12
Elimination of Unused Dependencies – SB-1509

rlevy12:
**Elimination of Unused Dependencies – SB-1509**
In the course of WAR validation, it was found that two dependencies needed to be removed from the Student and Proctor WARs. 
1. neo4j JAR (neo4j-cypher-compiler-2.1) is referenced in the student POM but is not referenced in the code. The result of this is its unnecessary inclusion of this JAR in the student WAR.  
2. mysql-connector JAR (mysql-connector-java) is a run-time dependency and should not be included in the WAR, but rather in the server's common.loader lib directory. This reference was removed from both the student and proctor POMs.


jtreuting:
Replaced by https://github.com/SmarterApp/TDS_Student/pull/10

PR #6, opened 2016-12-12
Intelligent Muting Capability – SB-1506

rlevy12:
**Intelligent Muting Capability – SB-1506**
Intelligent muting is a configuration (test level or proctor level) that allows the ability to (a) mute the entire test; (b) mute just passages; (c) mute just questions; or (d) don't mute anything. This was implemented with a new, custom tag (attribute) called brailleonly which is set to true for all item containers, or true for all passage containers, or both. This new attribute is only supported by JAWS 16+ and requires Braille mode to be active.
The brailleonly.js file that sets attribute brailleonly depends on the values of 'Mute System Volume' accommodation property derived from client_testtool table for given test.


jtreuting:
Replaced by https://github.com/SmarterApp/TDS_Student/pull/10

PR #11, opened 2017-01-21
Feature/open exam integration

millerts:


pom.xml
```diff
@@ -0,0 +1,22 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
millerts, line 1: This pom.xml builds iris, student, and student library.  I'll add the other projects as well.

student.library/pom.xml
```diff
@@ -45,6 +53,23 @@
 			<version>${item-scoring.version}</version>
 		</dependency>
 
+		<dependency>
+			<groupId>org.opentestsystem.common</groupId>
+			<artifactId>tds-common-legacy</artifactId>
+			<version>${tds-common-legacy.version}</version>
+		</dependency>
+
+		<dependency>
```
millerts, line 62: I will remove this dependency.  I was using it for message translation but am no longer doing that here.

student.library/src/main/java/tds/student/sql/abstractions/ExamRepository.java
```diff
@@ -0,0 +1,14 @@
...
+
+import tds.common.Response;
+import tds.exam.Exam;
+import tds.exam.OpenExamRequest;
+
+/**
+ * Repository to interact with exam data
+ */
+public interface ExamRepository {
+  Response<Exam> openExam(OpenExamRequest openExamRequest) throws ReturnStatusException;
```
millerts: I'll add the missing javadoc.

student/src/main/java/tds/student/configuration/WebConfiguration.java
```diff
@@ -0,0 +1,13 @@
...
+
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.web.client.RestTemplate;
+
+@Configuration
+public class WebConfiguration {
+    @Bean(name = "integrationRestTemplate")
+    public RestTemplate getRestTemplate() {
+        return new RestTemplate();
```
millerts, line 11: I did this because for now we don't need to use the AIR custom configured rest templates.  Depending on how we do our setup we may be able to continue to use the default.

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -0,0 +1,141 @@
...
+import tds.student.sql.data.OpportunityStatusChange;
+import tds.student.sql.data.OpportunityStatusExtensions;
+import tds.student.sql.data.TestConfig;
+import tds.student.sql.data.TestSegment;
+import tds.student.sql.data.TestSelection;
+import tds.student.sql.data.TestSession;
+import tds.student.sql.data.Testee;
+
+@Service("integrationOpportunityService")
+@Scope("prototype")
```
millerts, line 34: I'm not 100% sure what is happening here but with the way the application settings are loaded into the system by hooking into the Spring container forces us to use the prototype rather than the Spring default (singleton per container)

millerts: Also, properties only seem to work at the service layer and down.  The Web layer doesn't load the progman properties. :/

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -0,0 +1,141 @@
...
+  @Override
+  public List<TestSelection> getEligibleTests(Testee testee, TestSession session, String grade, BrowserInfo browserInfo) throws ReturnStatusException {
+    return legacyOpportunityService.getEligibleTests(testee, session, grade, browserInfo);
+  }
+
+  @Override
+  public OpportunityInfo openTest(Testee testee, TestSession session, String testKey) throws ReturnStatusException {
+    OpportunityInfo legacyOpportunityInfo = legacyOpportunityService.openTest(testee, session, testKey);
+
+    //This isn't ideal, but due to the way the progman properties are loaded within the system this lives within the service rather than the callers.
```
millerts, line 71: I'll remove this comment.  It was here before I figured out how to get the legacy service to be within this class.  This was solved via the prototype comment above.

student/src/main/java/tds/student/sql/repository/RemoteExamRepository.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.web.client.RestClientException;
+import org.springframework.web.client.RestTemplate;
+
+import tds.common.Response;
+import tds.exam.Exam;
+import tds.exam.OpenExamRequest;
+import tds.student.sql.abstractions.ExamRepository;
+
+@Repository
+public class RemoteExamRepository implements ExamRepository {
```
millerts, line 21: I go back and forth whether we need a repository but I feel it is cleaner to have our data access be here rather than have the service directly make the http calls.

student/src/main/java/tds/student/web/handlers/MasterShellHandler.java
```diff
@@ -141,6 +137,10 @@
   @Autowired
   private ItemBankService         itemBankService;
 
+  @Autowired
+  @Qualifier("integrationOpportunityService")
```
millerts, line 141: This is our remote opportunity service.  It still calls the legacy one in addition to making the http call.

student.library/src/main/java/tds/student/sql/data/OpportunityStatusExtensions.java
```diff
@@ -9,12 +9,16 @@
...
+import java.util.Map;
 
 import org.apache.commons.lang3.StringUtils;
 
+import tds.exam.ExamStatusCode;
+
 public class OpportunityStatusExtensions
 {
-  private static final HashMap<String, OpportunityStatusType> _statuses = new HashMap<String, OpportunityStatusType> ();
+  private static final Map<String, OpportunityStatusType> _statuses = new HashMap<String, OpportunityStatusType> ();
```
jtreuting: I know we aren't there yet, but in theory this could go away at some point in the future right?

student/pom.xml
```diff
@@ -47,7 +49,17 @@
...
+			<groupId>org.opentestsystem.exam</groupId>
+			<artifactId>tds-exam-client</artifactId>
+			<version>${tds-exam-client.version}</version>
+		</dependency>
 
+		<dependency>
+			<groupId>org.opentestsystem.common</groupId>
+			<artifactId>tds-common-legacy</artifactId>
+			<version>${tds-common-legacy.version}</version>
+		</dependency>
```
jtreuting, line 62: You mention above this wasn't needed in the parent pom anymore because of where the message translation moved.  Is this one needed?

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -0,0 +1,141 @@
...
+  }
+
+  @Override
+  public List<TestSelection> getEligibleTests(Testee testee, TestSession session, String grade, BrowserInfo browserInfo) throws ReturnStatusException {
+    return legacyOpportunityService.getEligibleTests(testee, session, grade, browserInfo);
+  }
+
+  @Override
+  public OpportunityInfo openTest(Testee testee, TestSession session, String testKey) throws ReturnStatusException {
+    OpportunityInfo legacyOpportunityInfo = legacyOpportunityService.openTest(testee, session, testKey);
```
jtreuting: I'd like to be able to do load testing and just call the new code easily so we can check some status along the way.  Let's add a flag that controls whether the legacy is called.  Maybe it is tri-state: call both, call legacy only, call new only, or 2 separate flags.

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -0,0 +1,141 @@
...
+      .withSessionId(session.getKey())
+      .withBrowserId(UUID.randomUUID())
+      .withStudentId(testee.getKey())
+      .build();
+
+    Response<Exam> response = examRepository.openExam(openExamRequest);
+
+    if (!response.hasError() && !response.getData().isPresent()) {
+      throw new ReturnStatusException("Invalid response from the exam service");
+    } else if (response.getError().isPresent()) {
```
jtreuting: Don't need the `else` since this throws.

student/src/main/java/tds/student/sql/repository/RemoteExamRepository.java
jtreuting: I agree it is cleaner.

jtreuting (changes requested):
looks good just one suggestion so we can switch modes between legacy, new or both

pom.xml
millerts: Done

student.library/pom.xml
millerts: Done

student/pom.xml
millerts: The dependency I was referencing was the auto value dependency. This is needed because of the Response object in the common legacy project.

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
millerts: Fixed.

burzum619 (approved):

PR #13, opened 2017-02-01
Feature/exam id in cookie

millerts:
Contains two main things:
1. Contains the necessary modules configured into the converter for the rest template to be able to parse the json from the microservices 
2. We needed to be additive in how we store data in the cookie so the legacy system can still work and we also have enough data to make the subsequent remote calls.

### DIFF
--- a/student.library/src/main/java/tds/student/sql/data/OpportunityInfo.java
+++ b/student.library/src/main/java/tds/student/sql/data/OpportunityInfo.java
@@ -1,7 +1,7 @@
 /*******************************************************************************
  * Educational Online Test Delivery System 
  * Copyright (c) 2014 American Institutes for Research
- *     
+ *
  * Distributed under the AIR Open Source License, Version 1.0 
  * See accompanying file AIR-License-1_0.txt or at
  * http://www.smarterapp.org/documents/American_Institutes_for_Research_Open_Source_Software_License.pdf
@@ -14,19 +14,27 @@ import com.fasterxml.jackson.annotation.JsonProperty;
 
 /**
  * @author temp_rreddy
- * 
+ *
  */
 public class OpportunityInfo
 {
   private OpportunityStatusType _status;
   private UUID                  _browserKey;
   private UUID                  _oppKey;
+  private UUID                  examBrowserKey;
+  private UUID                  examId;
+  private OpportunityStatusType examStatus;
 
   @JsonProperty ("IsOpen")
   public boolean getIsOpen () {
 
     return (_status == OpportunityStatusType.Pending || _status == OpportunityStatusType.Suspended || _status == OpportunityStatusType.Approved);
 
+  }
+
+  @JsonProperty ("IsExamOpen")
+  public boolean getIsExamOpen () {
+    return (examStatus == OpportunityStatusType.Pending || examStatus == OpportunityStatusType.Suspended || examStatus == OpportunityStatusType.Approved);
   }
 
   public OpportunityInfo () {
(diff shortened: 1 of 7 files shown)


student.library/src/main/java/tds/student/sql/data/OpportunityInfo.java
```diff
@@ -29,6 +32,11 @@ public boolean getIsOpen () {
 
   }
 
+  @JsonProperty ("IsExamOpen")
```
jtreuting, line 35: I'm used to json properties being `isExamOpen`, is there a standard in the code now we are following?

burzum619 (approved):

student.library/src/main/java/tds/student/sql/data/OpportunityInstance.java
```diff
@@ -37,9 +39,25 @@ public UUID getBrowserKey () {
     return _browserKey;
   }
 
+  @JsonProperty("ExamId")
```
burzum619, line 42: Curious, why the Title Casing here?

millerts: I am following the existing style.  The current code base uses Title Casing so I carried it forward in this class.

student/src/main/java/tds/student/web/StudentContext.java
```diff
@@ -188,20 +193,24 @@ public static String getGrade () {
...
-    // check if opp instance exists
-    // if (oppKey == Guid.Empty || sessionKey == Guid.Empty || browserKey ==
-    // Guid.Empty) return null;
-    if (oppKey == null || sessionKey == null || browserKey == null)
-      return null;
+    UUID examBrowserKey = StudentCookie.getCookieData(EXAM_BROWSER_KEY) == null
+      ? null
+      : UUID.fromString(StudentCookie.getCookieData(EXAM_BROWSER_KEY));
+
+    UUID examId = tdsIdentity.get(EXAM_BROWSER_KEY) == null
```
jtreuting: Is `EXAM_BROWSER_KEY` supposed to be for both of these?  `examBrowserKey` and `examId` . Maybe it should be `EXAM_ID_COOKIE_KEY` for one of them?

jtreuting (commented):
Clarifying question, might need to change, not sure.

student.library/src/main/java/tds/student/sql/data/OpportunityInstance.java
jtreuting: Ooops, just saw this.  Ignore my comment on the same thing later on

student/src/main/java/tds/student/web/StudentContext.java
millerts: Oops.  Bit by a late refactor.  Fixed.

student.library/src/main/java/tds/student/sql/data/OpportunityInfo.java
```diff
@@ -63,12 +71,42 @@ public void setOppKey (UUID _oppKey) {
...
+  @JsonProperty ("ExamId")
+  public UUID getExamId() {
+    return examId;
+  }
+
+  public void setExamId(UUID examId) {
+    this.examId = examId;
+  }
+
+  public void setExamStatus(OpportunityStatusType _status) {
```
jeffjohnson9046, line 92: minor:  `ExamStatus` setter before `ExamStatus` getter.

jeffjohnson9046 (approved):

PR #15, opened 2017-02-06
Handle the Response object with Validation errors

millerts:


student/src/main/java/tds/student/sql/repository/RemoteExamRepository.java
```diff
@@ -21,30 +27,54 @@
...
+      ResponseEntity<Response<Exam>> responseEntity = restTemplate.exchange(
         examUrl,
         HttpMethod.POST,
         requestHttpEntity,
         new ParameterizedTypeReference<Response<Exam>>() {
         });
+
+      response = responseEntity.getBody();
+    } catch (HttpClientErrorException hce) {
+      if(isClientError(hce.getStatusCode())) {
```
millerts, line 56: When rest template gets a 400 it throws.  When that happens we can't get to the body to get the Response object containing the `ValidationError`.  Spring jira [here](https://jira.spring.io/browse/SPR-10961).

We have two options.  We can either do what we have here or register our own error handler that will throw a unique exception when it is a 400 rather than a 500.  Now that I've written this code that may be what we want to do going forward but I figured I'd throw it to the team.

jtreuting (approved):
Makes sense to me.  Looks good.

PR #16, opened 2017-02-06
Integration of exam approval (for practice tests/guests)

burzum619:
This PR includes code in TDS_Student for calling the new ExamService approve accommodations endpoint. This endpoint will be called when an guest user needs approval for additional/selectable exam accommodations.

### DIFF
--- a/student.library/src/main/java/tds/student/sql/abstractions/ExamRepository.java
+++ b/student.library/src/main/java/tds/student/sql/abstractions/ExamRepository.java
@@ -2,7 +2,10 @@ package tds.student.sql.abstractions;
 
 import TDS.Shared.Exceptions.ReturnStatusException;
 
+import java.util.UUID;
+
 import tds.common.Response;
+import tds.exam.ApproveAccommodationsRequest;
 import tds.exam.Exam;
 import tds.exam.OpenExamRequest;
 
@@ -11,4 +14,5 @@ import tds.exam.OpenExamRequest;
  */
 public interface ExamRepository {
   Response<Exam> openExam(final OpenExamRequest openExamRequest) throws ReturnStatusException;
+  void approveAccommodations(final UUID examId, final ApproveAccommodationsRequest approveAccommodationsRequest) throws ReturnStatusException;
 }
--- a/student.library/src/main/java/tds/student/sql/abstractions/ExamRepository.java
+++ b/student.library/src/main/java/tds/student/sql/abstractions/ExamRepository.java
@@ -14,5 +14,13 @@ import tds.exam.OpenExamRequest;
  */
 public interface ExamRepository {
   Response<Exam> openExam(final OpenExamRequest openExamRequest) throws ReturnStatusException;
+    
+  /**
+   * Creates a request for the exam service to approve {@link tds.exam.ExamAccommodation}s
+   *
+   * @param examId                       the id of the {@link tds.exam.Exam}
+   * @param approveAccommodationsRequest the {@link tds.exam.ApproveAccommodationsRequest} containing approval request data
+   * @throws ReturnStatusException
+   */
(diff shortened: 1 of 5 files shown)


student/src/main/java/tds/student/services/remote/RemoteAccommodationsService.java
```diff
@@ -0,0 +1,94 @@
...
+    examRepository.approveAccommodations(oppInstance.getExamId(), request);
+  }
+  
+  @Override
+  public List<Accommodations> getApproved(OpportunityInstance opportunityInstance, String testKey, boolean isGuestSession) throws ReturnStatusException {
+    return legacyAccommodationsService.getApproved(opportunityInstance, testKey, isGuestSession);
+  }
+  
+  
+  private Map<Integer, Set<String>> parseSegmentAccommodationStrings(List<String> segmentsAccommodationData) {
```
jtreuting, line 80: It would be nice to get some unit tests around this logic to make sure it handles edge cases okay

student.library/src/main/java/tds/student/sql/abstractions/ExamRepository.java
```diff
@@ -11,4 +14,5 @@
  */
 public interface ExamRepository {
   Response<Exam> openExam(final OpenExamRequest openExamRequest) throws ReturnStatusException;
+  void approveAccommodations(final UUID examId, final ApproveAccommodationsRequest approveAccommodationsRequest) throws ReturnStatusException;
```
millerts: Can we add JavaDocs to the new method?  I created one for the open exam on the feature branch

student/src/main/java/tds/student/services/remote/RemoteAccommodationsService.java
```diff
@@ -0,0 +1,94 @@
...
+  }
+
+  @Override
+  public void approve(OpportunityInstance oppInstance, List<String> segmentsAccommodationData) throws ReturnStatusException {
+
+    if (isLegacyCallsEnabled) {
+      legacyAccommodationsService.approve(oppInstance, segmentsAccommodationData);
+    }
+
+    if (!isRemoteExamCallsEnabled || segmentsAccommodationData == null) {
```
millerts: Why are we bailing from the remote call if the segmentsAccommodationData is null even though we call the legacy code?  It seems that the segmentsAccommodation being null should short circuit it all?

If not, can we add comments as to why it is like this?

student/src/main/java/tds/student/sql/repository/RemoteExamRepository.java
```diff
@@ -47,4 +56,25 @@ public RemoteExamRepository(@Qualifier("integrationRestTemplate") final RestTemp
...
+    HttpHeaders headers = new HttpHeaders();
+    headers.set("Accept", MediaType.APPLICATION_JSON_VALUE);
+    headers.setContentType(MediaType.APPLICATION_JSON);
+    HttpEntity<?> requestHttpEntity = new HttpEntity<>(approveAccommodationsRequest, headers);
+
+    UriComponentsBuilder builder = UriComponentsBuilder.fromHttpUrl(String.format("%s/%s/accommodations", examUrl, examId));
+
+    try {
+      restTemplate.exchange(
+        builder.build().encode().toUri(),
```
millerts: The encode is unnecessary

student/src/main/java/tds/student/services/remote/RemoteAccommodationsService.java
burzum619: The legacy code also bails out early if segmentData == null;  (line 275,275 in AccommodationService.java) Ultimately I guess we can just bail out immediately if segmentData == null, and not call either service. Strange that it doesnt result in any sort of warning or error.

PR #17, opened 2017-02-10
/checkApproval integration

burzum619:
This PR includes implementation of /checkApproval, which serves two primary purposes:

1. Retrieves and validates the current status of the exam and
2. Retrieves all eligible exam accommodations

The accommodation metadata present in Accommodation/ExamAccommodations is then mapped to a legacy "Accommodations" class. This class is then used for creating the TDS-Student-Accs cookie, as well as the accommodation data returned to the client. 

There is some code in Assessment and Exam services that will accompany this PR. In particular, a list of accommodation Dependencies is now included in the assessment object.

### DIFF
--- a/student/src/main/java/tds/student/services/remote/RemoteAccommodationsService.java
+++ b/student/src/main/java/tds/student/services/remote/RemoteAccommodationsService.java
@@ -16,7 +16,7 @@ import java.util.Map;
 import java.util.Set;
 
 import tds.accommodation.Accommodation;
-import tds.accommodation.Dependency;
+import tds.accommodation.AccommodationDependency;
 import tds.assessment.Assessment;
 import tds.assessment.Segment;
 import tds.exam.ApproveAccommodationsRequest;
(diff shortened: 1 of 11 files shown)


student/src/main/java/tds/student/services/remote/RemoteAccommodationsService.java
```diff
@@ -27,55 +34,150 @@
...
+        // If there are no Accommodations created for this segment position, we need to create one.
+        if (retAccommodations == null) {
+          // Accommodation context is either the segmentId or assessmentId, depending on whether the accommodation is
+          // segment specific or global to the assessment
+          retAccommodations = new Accommodations(segmentPosition, accommodation.getContext(), getLabelForAssessmentOrSegment(segmentPosition, assessment));
+          segmentPositionToAccommodations.put(segmentPosition, retAccommodations);
+        }
+      
+        /* AccommodationService [399-400] - If this isn't a guest session, then no need to check isDisabledOnGuest flag */
+        if (!isGuestSession || (isGuestSession && !accommodation.isDisableOnGuestSession())) {
```
jtreuting: the `isGuestSession` in the OR part is not needed since if it gets there it means it is a guest session

jtreuting (approved):
1 small thing, but not a deal breaker at all.  looks good

PR #18, opened 2017-02-15
/denyApproval and /setStatus implementations

burzum619:
This PR includes implementations of the /setStatus and /denyApproval integration with the performance exam service. The legacy implementations of these methods are found in OpportunityService.java. 

There are a few minor changes to other repos that this PR will depend on:

TDS_Common - NoContentResponseResource.java 
  - Added private empty constructor for jackson deserialization
TDS_Exam - ExamStatusCode and StatusTransitionValidator
  - Made status transition and state map lookups to be case-insensitive.

### DIFF
--- a/student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
+++ b/student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
@@ -171,23 +171,23 @@ public class RemoteOpportunityService implements IOpportunityService {
     if (!statusChange.isCheckReturnStatus()) {
       return true;
     }
-  
-    if (maybeError.isPresent()) {
-      ValidationError error = maybeError.get();
-      returnStatus = new ReturnStatus(error.getCode(), error.getMessage());
-      
-      if (ExamStatusCode.STATUS_FAILED.equalsIgnoreCase (maybeError.get().getCode())) {
-        throw new ReturnStatusException (returnStatus);
-      }
-      /* We can skip line [234-237] as trying to update to an invalid status should result in a ValidationError generated
-         by the ExamService. */
-      
-      log.warn("Error setting exam status for exam id {}: Failed to set status to '{}' - {}",
-        oppInstance.getExamId(), statusChange.getStatus(), returnStatus.getReason());
-      isApproved = false;
-    } else {
-      isApproved = true;
+    
+    if (!maybeError.isPresent()) {
+      return true;
     }
+  
+    ValidationError error = maybeError.get();
+    returnStatus = new ReturnStatus(error.getCode(), error.getMessage());
+    
+    if (ExamStatusCode.STATUS_FAILED.equalsIgnoreCase (maybeError.get().getCode())) {
+      throw new ReturnStatusException (returnStatus);
+    }
+    /* OpportunityService - We can skip line [234-237] as trying to update to an invalid status should result in a ValidationError generated
+       by the ExamService. */
+    
+    log.warn("Error setting exam status for exam id {}: Failed to set status to '{}' - {}",
+      oppInstance.getExamId(), statusChange.getStatus(), returnStatus.getReason());
+    isApproved = false;
     
     return isApproved;
   }
@@ -225,7 +225,7 @@ public class RemoteOpportunityService implements IOpportunityService {
     
     OpportunityStatus opportunityStatus = getStatus(oppInstance);
   
-    /* Conditional on line [257] */
+    /* OpportunityService - Conditional on line [257] */
     if (opportunityStatus.getStatus() == OpportunityStatusType.Paused) {
       return;
     }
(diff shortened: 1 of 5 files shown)


student/src/main/java/tds/student/sql/repository/RemoteExamRepository.java
```diff
@@ -125,11 +136,11 @@ private static boolean isClientError(HttpStatus status) {
     
     try {
       responseEntity = restTemplate.exchange(
-          builder.build().encode().toUri(),
```
burzum619, line 128: My apologies for these accidental formatting commits - but on that note, are we using 2-space indents or 4-space indents for new lines in legacy TDS_Student?

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -173,7 +214,23 @@ public ApprovalInfo checkSegmentApproval(final OpportunityInstance oppInstance)
...
+      legacyOpportunityService.denyApproval(oppInstance);
+    }
+
+    if (!isRemoteExamCallsEnabled) {
+      return;
+    }
+    
+    OpportunityStatus opportunityStatus = getStatus(oppInstance);
+  
+    /* Conditional on line [257] */
```
millerts: Can you add the class name?  I understand it is in the legacy opportunity service but the comment will make more sense for people that aren't in the know.

millerts: Same for anywhere we reference line numbers.

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -148,7 +155,41 @@ public OpportunityStatus getStatus(final OpportunityInstance oppInstance) throws
...
+        throw new ReturnStatusException (returnStatus);
+      }
+      /* We can skip line [234-237] as trying to update to an invalid status should result in a ValidationError generated
+         by the ExamService. */
+      
+      log.warn("Error setting exam status for exam id {}: Failed to set status to '{}' - {}",
+        oppInstance.getExamId(), statusChange.getStatus(), returnStatus.getReason());
+      isApproved = false;
+    } else {
+      isApproved = true;
```
millerts: Can we flip this conditional so it is:

```
if (!maybeError.isPresent()) {
  return true;
}

Validation error = maybeError.get();
//rest of logic

return false; // if not excpetion thrown
```

student/src/main/java/tds/student/sql/repository/RemoteExamRepository.java
```diff
@@ -96,16 +107,16 @@ private static boolean isClientError(HttpStatus status) {
...
+      .queryParam("browserId", browserId);
     
     try {
       responseEntity = restTemplate.exchange(
-          builder.build().encode().toUri(),
-          HttpMethod.GET,
-          requestHttpEntity,
-          new ParameterizedTypeReference<Response<ExamApproval>>() {
-          });
+        builder.build().encode().toUri(),
```
millerts, line 115: I must have had the performance code style setting when I changed this code.

student/src/main/java/tds/student/sql/repository/RemoteExamRepository.java
```diff
@@ -157,4 +168,39 @@ public void approveAccommodations(UUID examId, ApproveAccommodationsRequest appr
...
+  
+  @Override
+  public Optional<ValidationError> updateStatus(final UUID examId, final String status, final String reason) throws ReturnStatusException {
+    HttpHeaders headers = new HttpHeaders();
+    headers.set("Accept", MediaType.APPLICATION_JSON_VALUE);
+    headers.setContentType(MediaType.APPLICATION_JSON);
+    HttpEntity<?> requestHttpEntity = new HttpEntity<>(headers);
+    
+    UriComponentsBuilder builder = UriComponentsBuilder.fromHttpUrl(String.format("%s/%s/status", examUrl, examId))
+      .queryParam("status", status)
```
millerts, line 180: This is nothing to do with your code but it seems odd to me that we are using query parameters rather than putting a body object with these values.  Query params to me are normally used to impact the query being done.  Like search parameters.

student/src/main/java/tds/student/sql/repository/RemoteExamRepository.java
```diff
@@ -87,6 +89,15 @@ private static boolean isClientError(HttpStatus status) {
     }
   }
   
+  private NoContentResponseResource handleErrorResponseNoContent(String body) throws ReturnStatusException {
```
millerts, line 92: Can you add a unit test for that hits this block?

millerts (changes requested):
Couple small requests.

student/src/main/java/tds/student/sql/repository/RemoteExamRepository.java
millerts: 2 space because that is what they use.  I imagine I didn't have the write code style in my IDE and missed it.

student/src/main/java/tds/student/sql/repository/RemoteExamRepository.java
burzum619: Agreed. we can chat about it in the parking lot tomorrow.

PR #20, opened 2017-02-17
Start Exam Integration

burzum619:
Integration of start exam endpoint to legacy student. This code corresponds to OpportunityService.startTest(). Primary entrypoint is RemoteOpportunityService.startTest(), which is called by the MasterShellHandler.java

### DIFF
--- a/student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
+++ b/student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
@@ -20,6 +20,7 @@ import tds.common.ValidationError;
 import tds.exam.Exam;
 import tds.exam.ExamApproval;
 import tds.exam.ExamConfiguration;
+import tds.exam.ExamSegment;
 import tds.exam.ExamStatusCode;
 import tds.exam.OpenExamRequest;
 import tds.student.services.abstractions.IOpportunityService;
@@ -54,7 +55,7 @@ public class RemoteOpportunityService implements IOpportunityService {
     @Value("${tds.exam.legacy.enabled}") Boolean legacyCallsEnabled,
     ExamRepository examRepository) {
 
-    if(!remoteExamCallsEnabled && !legacyCallsEnabled) {
+    if (!remoteExamCallsEnabled && !legacyCallsEnabled) {
       throw new IllegalStateException("Remote and legacy calls are both disabled.  Please check progman configuration");
     }
 
@@ -73,7 +74,7 @@ public class RemoteOpportunityService implements IOpportunityService {
   public OpportunityInfo openTest(final Testee testee, final TestSession session, final String testKey) throws ReturnStatusException {
     OpportunityInfo opportunityInfo = new OpportunityInfo();
 
-    if(isLegacyCallsEnabled) {
+    if (isLegacyCallsEnabled) {
       opportunityInfo = legacyOpportunityService.openTest(testee, session, testKey);
     }
 
@@ -122,7 +123,7 @@ public class RemoteOpportunityService implements IOpportunityService {
   public OpportunityStatus getStatus(final OpportunityInstance oppInstance) throws ReturnStatusException {
     OpportunityStatus status = null;
 
-    if(isLegacyCallsEnabled) {
+    if (isLegacyCallsEnabled) {
       status = legacyOpportunityService.getStatus(oppInstance);
     }
 
@@ -132,7 +133,7 @@ public class RemoteOpportunityService implements IOpportunityService {
     }
 
     Response<ExamApproval> response = examRepository.getApproval(oppInstance.getExamId(), oppInstance.getSessionKey(),
-        oppInstance.getExamBrowserKey());
+      oppInstance.getExamBrowserKey());
 
     if (!response.hasError() && !response.getData().isPresent()) {
       throw new ReturnStatusException("Invalid response from the exam service");
@@ -141,8 +142,8 @@ public class RemoteOpportunityService implements IOpportunityService {
     if (response.getError().isPresent()) {
       ValidationError validationError = response.getError().get();
       String errorMessage = validationError.getTranslatedMessage().isPresent()
-          ? validationError.getTranslatedMessage().get()
-          : validationError.getMessage();
+        ? validationError.getTranslatedMessage().get()
+        : validationError.getMessage();
 
       throw new ReturnStatusException(errorMessage);
     }
@@ -158,38 +159,38 @@ public class RemoteOpportunityService implements IOpportunityService {
   public boolean setStatus(final OpportunityInstance oppInstance, final OpportunityStatusChange statusChange) throws ReturnStatusException {
     boolean isApproved = false;
     ReturnStatus returnStatus = null;
-    
+
     if (isLegacyCallsEnabled) {
       isApproved = legacyOpportunityService.setStatus(oppInstance, statusChange);
     }
-  
+
     if (!isRemoteExamCallsEnabled) {
       return isApproved;
     }
-    
+
     Optional<ValidationError> maybeError = examRepository.updateStatus(oppInstance.getExamId(), statusChange.getStatus().name(), statusChange.getReason());
-    
+
     if (!statusChange.isCheckReturnStatus()) {
       return true;
     }
-    
+
     if (!maybeError.isPresent()) {
       return true;
     }
-  
+
     ValidationError error = maybeError.get();
     returnStatus = new ReturnStatus(error.getCode(), error.getMessage());
-    
-    if (ExamStatusCode.STATUS_FAILED.equalsIgnoreCase (maybeError.get().getCode())) {
-      throw new ReturnStatusException (returnStatus);
+
+    if (ExamStatusCode.STATUS_FAILED.equalsIgnoreCase(maybeError.get().getCode())) {
+      throw new ReturnStatusException(returnStatus);
     }
     /* OpportunityService - We can skip line [234-237] as trying to update to an invalid status should result in a ValidationError generated
        by the ExamService. */
-    
+
     log.warn("Error setting exam status for exam id {}: Failed to set status to '{}' - {}",
       oppInstance.getExamId(), statusChange.getStatus(), returnStatus.getReason());
     isApproved = false;
-    
+
     return isApproved;
   }
 
@@ -197,7 +198,7 @@ public class RemoteOpportunityService implements IOpportunityService {
   public ApprovalInfo checkTestApproval(final OpportunityInstance oppInstance) throws ReturnStatusException {
     ApprovalInfo approvalInfo = null;
 
-    if(isLegacyCallsEnabled) {
+    if (isLegacyCallsEnabled) {
       approvalInfo = legacyOpportunityService.checkTestApproval(oppInstance);
     }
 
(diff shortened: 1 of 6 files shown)


student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -235,7 +236,43 @@ public void denyApproval(final OpportunityInstance oppInstance) throws ReturnSta
...
+  
+    
+    /* Note that the formKeys argument can be ignored - it is an unused functionality */
+    Response<ExamConfiguration> response = examRepository.startExam(oppInstance.getExamId());
+  
+    if (response.getError().isPresent()) {
+      ValidationError validationError = response.getError().get();
+      String errorMessage = validationError.getTranslatedMessage().isPresent()
+        ? validationError.getTranslatedMessage().get()
+        : validationError.getMessage();
```
wtritch: super-nit: these variables can have `final`

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -235,7 +236,43 @@ public void denyApproval(final OpportunityInstance oppInstance) throws ReturnSta
...
+    
+      throw new ReturnStatusException(errorMessage);
+    }
+    
+    if (!response.getData().isPresent()) {
+      throw new ReturnStatusException(String.format("Invalid response from the exam service when trying to start exam %s", oppInstance.getExamId()));
+    }
```
wtritch: Should this exception include the actual invalid response payload?  Or is that logged elsewhere?

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -235,7 +236,43 @@ public void denyApproval(final OpportunityInstance oppInstance) throws ReturnSta
...
+
+    ExamConfiguration examConfiguration = response.getData().get();
```
wtritch: super-nit: final (I'll stop commenting on finals now. =)

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -235,7 +236,43 @@ public void denyApproval(final OpportunityInstance oppInstance) throws ReturnSta
...
+    
+    /* OpportunityService - Conditional at line 288 */
+    if (!examConfiguration.getStatus().equals(ExamStatusCode.STATUS_STARTED)) {
+      throw new ReturnStatusException("Failed to start the exam.");
+    }
```
wtritch: Should this exception include the returned status code?

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -252,4 +289,22 @@ public void waitForSegment(final OpportunityInstance oppInstance, final int segm
...
+    testConfig.setRestart(exam.getRestartsAndResumptions());
+    testConfig.setTestLength(examConfiguration.getTestLength());
+    testConfig.setStartPosition(examConfiguration.getStartPosition());
+    testConfig.setContentLoadTimeout(examConfiguration.getContentLoadTimeoutMinutes());
+    testConfig.setInterfaceTimeout(examConfiguration.getInterfaceTimeoutMinutes());
+    testConfig.setRequestInterfaceTimeout(examConfiguration.getRequestInterfaceTimeoutMinutes());
+    testConfig.setOppRestartMins(examConfiguration.getExamRestartWindowMinutes());
+    testConfig.setPrefetch(examConfiguration.getPrefetch());
+    testConfig.setScoreByTDS(false);
+    testConfig.setValidateCompleteness(examConfiguration.isValidateCompleteness());
```
wtritch, line 327: whoo boy.  Lotta properties.  Is this the only place we map an ExamConfiguration to a TestConfig or should we have a TestConfigBuilder that can copy an ExamConfiguration?

student/src/main/java/tds/student/sql/repository/RemoteExamRepository.java
```diff
@@ -98,6 +99,15 @@ private NoContentResponseResource handleErrorResponseNoContent(String body) thro
     }
   }
   
+  private Response<ExamConfiguration> handleErrorResponseExamConfig(String body) throws ReturnStatusException {
+    try {
+      JavaType type = objectMapper.getTypeFactory().constructParametricType(Response.class, Exam.class);
+      return objectMapper.readValue(body, type);
+    } catch (IOException e) {
+      throw new ReturnStatusException(e);
```
wtritch, line 108: Could/should we include the body as a property in the ReturnStatusException?  We were running into debugging issues on my last project, where we weren't able to deserialize the payload but we didn't know why because we couldn't see the payload.

student/src/main/java/tds/student/sql/repository/RemoteExamRepository.java
```diff
@@ -203,4 +213,35 @@ public void approveAccommodations(UUID examId, ApproveAccommodationsRequest appr
...
   }
+  
+  @Override
+  public Response<ExamConfiguration> startExam(UUID examId) throws ReturnStatusException {
+    HttpHeaders headers = new HttpHeaders();
+    headers.set("Accept", MediaType.APPLICATION_JSON_VALUE);
+    headers.setContentType(MediaType.APPLICATION_JSON);
+    HttpEntity<?> requestHttpEntity = new HttpEntity<>(headers);
+    Response<ExamConfiguration> response;
+    UriComponentsBuilder builder = UriComponentsBuilder.fromHttpUrl(String.format("%s/%s/start", examUrl, examId));
```
wtritch, line 225: Just throwing it out there, but String.format can do funky things in different locales... (especially with numbers)

student/src/main/java/tds/student/sql/repository/RemoteExamRepository.java
```diff
@@ -203,4 +213,35 @@ public void approveAccommodations(UUID examId, ApproveAccommodationsRequest appr
...
+  
+    try {
+      ResponseEntity<Response<ExamConfiguration>> responseEntity = restTemplate.exchange(
+        builder.build().toUri(),
+        HttpMethod.PUT,
+        requestHttpEntity,
+        new ParameterizedTypeReference<Response<ExamConfiguration>>() {
+        });
```
wtritch, line 233: Have we looked into using Spring Retry for exponential-backoff retries on failed server-to-server communications? It worked well for us with 5xx "hiccups" (4xx errors were considered idempotent, so we rejected the request without retrying)

jtreuting: I like that idea.

student/src/main/java/tds/student/sql/repository/RemoteExamRepository.java
millerts: That's probably a good idea.  I didn't do that in the other place we have similar code and should've done that.

student/src/test/java/tds/student/services/remote/RemoteOpportunityServiceTest.java
```diff
@@ -283,4 +286,88 @@ public void shouldReturnFalseForValidationErrorReturned() throws ReturnStatusExc
...
+    service.startTest(oppInstance, assessmentKey, null);
+  }
+  
+  @Test(expected = ReturnStatusException.class)
+  public void shouldThrowForReturnStatusNotStarted() throws ReturnStatusException {
+    service = new RemoteOpportunityService(legacyOpportunityService, true, false, examRepository);
+    OpportunityInstance oppInstance = new OpportunityInstance(UUID.randomUUID(), UUID.randomUUID(), UUID.randomUUID());
+    final String assessmentKey = "assessmentKey";
+  
+    ExamConfiguration mockExamConfig = new ExamConfiguration.Builder()
```
millerts: Not a huge deal to me, but I believe you could simply do `new ExamConfiguration.Builder().withStatus(ExamStatusCode.STATUS_DENIED).build()` and the test would still pass.

Unrelated to this change but I'm surprised the status is fetched from the configuration rather than the exam.

student/src/test/java/tds/student/services/remote/RemoteOpportunityServiceTest.java
```diff
@@ -283,4 +286,88 @@ public void shouldReturnFalseForValidationErrorReturned() throws ReturnStatusExc
...
+    assertThat(testConfig.getPrefetch()).isEqualTo(mockExamConfig.getPrefetch());
+    assertThat(testConfig.getContentLoadTimeout()).isEqualTo(mockExamConfig.getContentLoadTimeoutMinutes());
+    assertThat(testConfig.getInterfaceTimeout()).isEqualTo(mockExamConfig.getInterfaceTimeoutMinutes());
+    assertThat(testConfig.getOppRestartMins()).isEqualTo(mockExamConfig.getExamRestartWindowMinutes());
+    assertThat(testConfig.getRequestInterfaceTimeout()).isEqualTo(mockExamConfig.getRequestInterfaceTimeoutMinutes());
+    assertThat(testConfig.getRestart()).isEqualTo(mockExamConfig.getExam().getRestartsAndResumptions());
+    assertThat(testConfig.getStartPosition()).isEqualTo(mockExamConfig.getStartPosition());
+  }
+}
+
```
millerts: Can we remove the extra line breaks?

student/src/main/java/tds/student/sql/repository/RemoteExamRepository.java
millerts: It probably isn't a huge issue for us today, but I agree we should probably instead use the [buildAndExpand](http://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/web/util/UriComponentsBuilder.html#buildAndExpand-java.lang.Object...-) feature.

millerts (approved):

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
burzum619: I guess its up to debate, but this ReturnStatusException actually gets displayed in the UI as an error and I'm not sure we want to include code-level details (which the exam status would be). Would be a prime candidate for error logging though. I can be convinced otherwise though

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
burzum619: Yep, I believe this is the only place we have to map these, which is why I kept it as a private static method in-class.

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
millerts: Did we check that there isn't a translated message that happens when it fails to start?

student/src/main/java/tds/student/sql/repository/RemoteExamRepository.java
millerts: I agree it is a good idea.  However, I don't think implementing that needs to be part of this PR.

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
burzum619: I have not checked if theres a translated message - how would I go about doing that?

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -252,4 +316,41 @@ public void waitForSegment(final OpportunityInstance oppInstance, final int segm
...
+    //TODO: set MSB flag
+    return testConfig;
+  }
+
+  private OpportunitySegment.OpportunitySegments mapExamSegmentsToOpportunitySegments(List<ExamSegment> examSegments) {
```
millerts: This can be static

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -252,4 +316,41 @@ public void waitForSegment(final OpportunityInstance oppInstance, final int segm
...
+    OpportunitySegment.OpportunitySegments opportunitySegments = new OpportunitySegment().new OpportunitySegments();
```
burzum619: I DO NOT like having to instantiate this silly inner-list class in this manner, but I could not find an alternative to doing it like this. Any suggestions are welcome.

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
burzum619: good catch

PR #21, opened 2017-02-23
Feature/exam opportunity id mapping

jtreuting:
Inserts a new record that maps examId to legacy testOpportunityId on openExam.  This data is then fetched on the Proctor side when the proctor UI calls the API with the new exam ids.  In order to also call the legacy services that exam id needs to be translated to the test opportunity id.

This is using the legacy performance way of doing things since it interacts with the legacy DB.  I created DAO and put it in the performance package.

I've implemented for approve, deny and approve accommodations on Proctor at this point.

NOTE: I'd like to discuss the best way to get the new table created.  I manually create in dev at this point and there isn't a migration path for the legacy tables that I'm aware of.

```
CREATE TABLE testopportunity_exam_map (
  testopportunity_id char(36) NOT NULL,
  exam_id char(36) NOT NULL,
  PRIMARY KEY (testopportunity_id,exam_id)
)
```

### DIFF
--- a/student/src/main/java/tds/student/performance/dao/TestOpportunityExamMapDao.java
+++ b/student/src/main/java/tds/student/performance/dao/TestOpportunityExamMapDao.java
@@ -3,5 +3,5 @@ package tds.student.performance.dao;
 import java.util.UUID;
 
 public interface TestOpportunityExamMapDao {
-    void insert(UUID testOpportunityId, UUID examId);
+    void insert(final UUID testOpportunityId, final UUID examId);
 }
--- a/student/src/main/java/tds/student/performance/dao/TestOpportunityExamMapDao.java
+++ b/student/src/main/java/tds/student/performance/dao/TestOpportunityExamMapDao.java
@@ -1,0 +1,7 @@
+package tds.student.performance.dao;
+
+import java.util.UUID;
+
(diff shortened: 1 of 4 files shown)


burzum619 (approved):
Looks good to me.

student/src/main/java/tds/student/performance/dao/TestOpportunityExamMapDao.java
```diff
@@ -0,0 +1,7 @@
+package tds.student.performance.dao;
+
+import java.util.UUID;
+
+public interface TestOpportunityExamMapDao {
```
millerts: I know the performance work didn't do this but can you add JavaDocs to the interface so future people know what this is doing.  When I joined the project in August I would've really appreciated JavaDocs on the interfaces.  Instead I had to look at implementation to figure out what was happening, and there were times even that was hard to discern intent.

millerts (approved):
Looks good to me.  To answer your `NOTE` if this is going to stay in the code when we release then I think we should leverage Flyway and do something like we do [here](https://github.com/SmarterApp/TDS_ExamService/blob/develop/service/pom.xml#L188).  

If we're planning to remove the flag and checks prior to May's release then I think manual creation is fine.

PR #22, opened 2017-02-24
Feature/remove unnecessary dependencies

millerts:
It'll be hard to see what is going on here.

* Created a new student-shared project for the stuff shared between student and iris
* Moved the shared classes between iris/student from student.library to student-shared
* Moved iris specific classes that were in student.library to iris.
* Remove any unused dependencies from the student.library jar
* Added dependencies to student that were missing and were being brought in via student.library


jtreuting (commented):
Looks like a lot of non-changed files came along for the ride

jtreuting (approved):
Just noticed these were renames of files and not formatting chagnes at it appears with the big bloc of red and gree

wtritch (commented):
I'm wondering if we actually want another sub-module, now that we're going to be releasing "client" libraries independently of the "service" and "parentPom" releases...

millerts:
@wtritch , I actually think that student.library shouldn't have existed in its previous state.  The student-shared module is the only thing that is actually shared across projects.  All the code in student.library could be within student.

Now that we are creating a scoring module it may be better to move the common shared scoring code into that module moving all the student stuff that isn't shared in the the student project.

That would result in:
* student.library going away
* student having the student only interfaces and classes as needed
* Scoring related classes and interfaces will be in `tds-scoring`.
* Student would have a dependency on `tds-scoring`

Not sure we're going to get to all of that in the current timelines but that is how I think it should be done.

I think if we decided to do what I list above we can create a separate PR for that work.

wtritch:
@millerts I haven't done a deep dive into the code, but I like what you're saying!  I agree that those changes are likely scope-creep on current tasking, however.

PR #25, opened 2017-03-03
Print/emboss integration

burzum619:
Implementation of the RemotePrintService, which will create ExamPrintRequests to send to the ExamService.

Please reference PrintServiceImpl (previously PrintService) for the prior implementation of this code.

I have renamed `PrintService` to `PrintServiceImpl` and created an interface `Print Service` which the (legacy) `PrintServiceImpl` and `RemotePrintService` now implements.

### DIFF
--- a/student.library/src/main/java/tds/student/services/abstractions/PrintService.java
+++ b/student.library/src/main/java/tds/student/services/abstractions/PrintService.java
@@ -1,0 +1,102 @@
+package tds.student.services.abstractions;
+
+import TDS.Shared.Exceptions.ReturnStatusException;
+
+import tds.itemrenderer.data.AccLookup;
+import tds.student.services.data.ItemResponse;
+import tds.student.services.data.PageGroup;
+import tds.student.services.data.TestOpportunity;
+import tds.student.sql.data.OpportunityInstance;
+
+/**
+ * A service for creating print requests
+ */
+public interface PrintService {
+  /**
+   * Creates a print request to print a passage
+   *
+   * @param oppInstance       The opportunity instance containing exam information
+   * @param pageGroupToPrint  The page group to print
+   * @param requestParameters Additional request parameters for the print request
+   * @return True if successful, false otherwise
+   * @throws ReturnStatusException
+   */
+  boolean printPassage(OpportunityInstance oppInstance, PageGroup pageGroupToPrint, String requestParameters) throws ReturnStatusException;
+
+  /**
+   * Creates a print request to print a page
+   *
+   * @param oppInstance       The opportunity instance containing exam information
+   * @param pageGroupToPrint  The page group to print
+   * @param requestParameters Additional request parameters for the print request
+   * @return True if successful, false otherwise
+   * @throws ReturnStatusException
+   */
+  boolean printPage(OpportunityInstance oppInstance, PageGroup pageGroupToPrint, String requestParameters) throws ReturnStatusException;
+
+  /**
+   * Creates a print request to print a passage or page
+   *
+   * @param oppInstance       The opportunity instance containing exam information
+   * @param pageGroupToPrint  The page group to print
+   * @param requestParameters Additional request parameters for the print request
+   * @return True if successful, false otherwise
+   * @throws ReturnStatusException
+   */
+  boolean printPassage(String requestType, OpportunityInstance oppInstance, PageGroup pageGroupToPrint, String requestParameters) throws ReturnStatusException;
+
+  /**
+   * Creates a print request to print an item
+   *
+   * @param oppInstance       The opportunity instance containing exam information
+   * @param responseToPrint   The item to print
+   * @param requestParameters Additional request parameters for the print request
+   * @return True if successful, false otherwise
+   * @throws ReturnStatusException
+   */
+  boolean printItem(OpportunityInstance oppInstance, ItemResponse responseToPrint, String requestParameters) throws ReturnStatusException;
+
+  /**
+   * Creates a print request to emboss a passage
+   *
+   * @param testOpp          The test opportunity object containing exam information
+   * @param pageGroupToPrint The page group to emboss
+   * @param accLookup        The map of exam accommodations
+   * @return True if successful, false otherwise
+   * @throws ReturnStatusException
+   */
+  boolean printPassageBraille(TestOpportunity testOpp, PageGroup pageGroupToPrint, AccLookup accLookup) throws ReturnStatusException;
+
+  /**
+   * Creates a print request to emboss a page
+   *
+   * @param testOpp          The test opportunity object containing exam information
+   * @param pageGroupToPrint The page group to emboss
+   * @param accLookup        The map of exam accommodations
+   * @return True if successful, false otherwise
+   * @throws ReturnStatusException
+   */
+  boolean printPageBraille(TestOpportunity testOpp, PageGroup pageGroupToPrint, AccLookup accLookup) throws ReturnStatusException;
+
+  /**
+   * Creates a print request to emboss a page or passage
+   *
+   * @param testOpp          The test opportunity object containing exam information
+   * @param pageGroupToPrint The page group to emboss
+   * @param accLookup        The map of exam accommodations
+   * @return True if successful, false otherwise
+   * @throws ReturnStatusException
+   */
+  boolean printPassageBraille(String requestType, TestOpportunity testOpp, PageGroup pageGroupToPrint, AccLookup accLookup) throws ReturnStatusException;
+
+  /**
+   * Creates a print request to emboss an item
+   *
+   * @param testOpp         The test opportunity object containing exam information
+   * @param responseToPrint The item to emboss
+   * @param accLookup       The map of exam accommodations
+   * @return True if successful, false otherwise
+   * @throws ReturnStatusException
+   */
+  boolean printItemBraille(TestOpportunity testOpp, ItemResponse responseToPrint, AccLookup accLookup) throws ReturnStatusException;
+}
--- a/student.library/src/main/java/tds/student/services/abstractions/PrintService.java
+++ b/student.library/src/main/java/tds/student/services/abstractions/PrintService.java
@@ -18,8 +18,8 @@ public interface PrintService {
    * @param oppInstance       The opportunity instance containing exam information
    * @param pageGroupToPrint  The page group to print
    * @param requestParameters Additional request parameters for the print request
-   * @return True if successful, false otherwise
-   * @throws ReturnStatusException
+   * @return {@code true} if successful, {@code false} otherwise
+   * @throws ReturnStatusException thrown if xml path for resource is null or empty
    */
   boolean printPassage(OpportunityInstance oppInstance, PageGroup pageGroupToPrint, String requestParameters) throws ReturnStatusException;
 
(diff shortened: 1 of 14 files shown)


student.library/src/main/java/tds/student/services/abstractions/PrintService.java
```diff
@@ -0,0 +1,102 @@
...
+  boolean printPassage(OpportunityInstance oppInstance, PageGroup pageGroupToPrint, String requestParameters) throws ReturnStatusException;
+
+  /**
+   * Creates a print request to print a page
+   *
+   * @param oppInstance       The opportunity instance containing exam information
+   * @param pageGroupToPrint  The page group to print
+   * @param requestParameters Additional request parameters for the print request
+   * @return True if successful, false otherwise
+   * @throws ReturnStatusException
```
millerts: If possible it'd be nice to mention in what scenario this is thrown.  Same for the other methods.  If it is a generic catch maybe something like `thrown if there is an issue`

student.library/src/main/java/tds/student/services/abstractions/PrintService.java
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * Creates a print request to emboss a page or passage
+   *
+   * @param testOpp          The test opportunity object containing exam information
+   * @param pageGroupToPrint The page group to emboss
+   * @param accLookup        The map of exam accommodations
+   * @return True if successful, false otherwise
+   * @throws ReturnStatusException
+   */
+  boolean printPassageBraille(String requestType, TestOpportunity testOpp, PageGroup pageGroupToPrint, AccLookup accLookup) throws ReturnStatusException;
```
millerts: The JavaDoc looks to be copy/paste from printPassageBraille above.  It is missing a doc for `requestType` and the description does not provide me with enough information to understand the difference between the two methods.

student.library/src/main/java/tds/student/services/abstractions/PrintService.java
```diff
@@ -0,0 +1,102 @@
...
+   */
+  boolean printPageBraille(TestOpportunity testOpp, PageGroup pageGroupToPrint, AccLookup accLookup) throws ReturnStatusException;
+
+  /**
+   * Creates a print request to emboss a page or passage
+   *
+   * @param testOpp          The test opportunity object containing exam information
+   * @param pageGroupToPrint The page group to emboss
+   * @param accLookup        The map of exam accommodations
+   * @return True if successful, false otherwise
```
millerts: minor nit, `{@code true}`

student/src/main/java/tds/student/services/PrintServiceImpl.java
```diff
@@ -178,10 +180,10 @@ public boolean printPassageBraille (String requestType, TestOpportunity testOpp,
       String xmlPath = pageGroupToPrint.getFilePath ();
       if (StringUtils.isEmpty (xmlPath)) {
         throw new ReturnStatusException (
-            new Exception (
-                String.format (
-                    "PrintPassageBraille: Invalid xml file path for group %1$s.",
-                    pageGroupToPrint.getId ())));
+          new Exception (
```
millerts: Can we get rid of the sub Exception?  The sub Exception adds no value and `ReturnStatusException` can take a String.  I understand you did not add this code but since we're changing it with the formatting we may as well improve it.

student/src/main/java/tds/student/services/remote/RemoteAccommodationsService.java
```diff
@@ -149,17 +149,29 @@ public void approve(OpportunityInstance oppInstance, List<String> segmentsAccomm
...
+
       for (AccommodationDependency assessmentDependency : assessment.getAccommodationDependencies()) {
         assessmentAccommodation.AddDependency(assessmentDependency.getIfType(), assessmentDependency.getIfValue(),
           assessmentDependency.getThenType(), assessmentDependency.getThenValue(), assessmentDependency.isDefault());
       }
     }
-    
+
+    // If this list does not contain the total number of segments + the parent assessment, we need to add a default Accommodations
+    if (segmentPositionToAccommodations.size() < assessment.getSegments().size() + 1) {
```
millerts, line 164: Can we flip this to be `if (assessment.getSegments().size() >= segmentPositionToAccommodations.size())`?  I think it is the same and we don't have to do the `+1`.

student/src/main/java/tds/student/services/remote/RemotePrintService.java
```diff
@@ -0,0 +1,360 @@
...
+      .withItemPosition(PASSAGE_PRINT_ITEM_POSITION_DEFAULT)
+      .withType(requestType)
+      .withDescription(requestDescription)
+      .withParameters(requestParameters)
+      .withValue(requestValue.replace("\\", "\\\\"))
+      .build();
+
+    examRepository.createPrintRequest(request);
+
+    isSuccessful = true;
```
millerts: What do you think about changing this to:

```
pageGroupToPrint.setPrinted(true);
return true;
```

It just reads easier to me.

student/src/main/java/tds/student/services/remote/RemotePrintService.java
```diff
@@ -0,0 +1,360 @@
...
+    // get attachments for the accommodations braille type
+    List<ITSAttachment> brailleAttachments = content.GetBrailleTypeAttachment(accLookup);
+
+    if (brailleAttachments.isEmpty()) { // no need for null check, this is initialized as an empty list
+      LOG.warn("PrintItemBraille: Cannot find a matching braille attachment for the test {} and item {}.",
+        testOpp.getTestKey(), responseToPrint.getItemID());
+      return false;
+    }
+
+    ITSAttachment mainBrailleAttachment = brailleAttachments.get(0);
```
millerts, line 317: Is there a way to extract a common method between this and the block in 207?  It is hard to pick the nuanced difference but this method and print passage braille seem have some overlap.

student/src/main/java/tds/student/services/remote/RemotePrintService.java
```diff
@@ -0,0 +1,360 @@
...
+  }
+
+  private String getPassageLabel(PageGroup group, boolean isTranscript) throws ReturnStatusException {
+    StringBuilder label = new StringBuilder("Passage");
+
+    if (isTranscript) {
+      label.append(" and Transcript");
+    }
+
+    if (group.size() > 0) {
```
millerts: Can we use `if(!group.isEmpty())`

student/src/main/java/tds/student/services/remote/RemotePrintService.java
```diff
@@ -0,0 +1,360 @@
...
+      label.append(" for ");
+
+      ItemResponse firstResponse = group.getFirst();
+      if (group.size() == 1) {
+        String postion = String.format("Item %1$d", firstResponse.getPosition());
+        label.append(postion);
+
```
millerts: Can we remove the extra line break?

student/src/main/java/tds/student/sql/repository/ExamRepository.java
```diff
@@ -89,4 +90,6 @@
    * @throws ReturnStatusException
    */
   Response<List<ExamSegment>> findExamSegments(final UUID examId, final UUID sessionId, final UUID browserId) throws ReturnStatusException;
+
+  void createPrintRequest(ExamPrintRequest examPrintRequest) throws ReturnStatusException;
```
millerts: Can we add a JavaDoc?

student/src/test/java/tds/student/services/PrintServiceTest.java
```diff
@@ -41,7 +41,7 @@
   private static final Logger _logger               = LoggerFactory.getLogger (PrintServiceTest.class);
 
   @Autowired
-  PrintService          printService ;
+  PrintServiceImpl printService ;
```
millerts: Can we try to use a qualifier here rather than changing it to the impl?

student/src/test/java/tds/student/services/remote/RemotePrintServiceTest.java
```diff
@@ -0,0 +1,599 @@
...
+
+  @Mock
+  private IContentService mockContentService;
+
+  private PrintService service;
+
+  @Captor
+  private ArgumentCaptor<ExamPrintRequest> examPrintRequestCaptor;
+
+  final OpportunityInstance mockOpportunityInstance = new OpportunityInstance(UUID.randomUUID(), UUID.randomUUID(), UUID.randomUUID());
```
millerts, line 54: I'm surprised you didn't use your random value generator library here

student/src/test/java/tds/student/web/handlers/ContentHandlerTest.java
```diff
@@ -53,7 +53,7 @@
   IContentService        _contentService;
 
   @Mock
-  PrintService           _printService;
+  PrintServiceImpl _printService;
```
millerts: Same comment about using the qualifier rather than using the impl

student/src/test/resources/test-context.xml
```diff
@@ -56,7 +56,7 @@
 	<!-- For ResponseRepositoryTest -->	
 	<beans:bean id="iResponseRepository" class="tds.student.sql.repository.ResponseRepository"/>
 	<!-- For PrintServiceTest -->
-	<beans:bean id="printService" class="tds.student.services.PrintService"/>
+	<beans:bean id="printService" class="tds.student.services.PrintServiceImpl"/>
```
millerts: Is this even needed?  Proctor is using a mix of XML and annotation based injection but it seems to me looking at the PrintServiceImpl class that the bean definition isn't needed?

millerts (approved):
Logic looks good.  Just a few comments.

student/src/test/java/tds/student/services/PrintServiceTest.java
burzum619: Nice catch, IntelliJ helpfully refactored this property datatype (as well as some of the other references that I actually caught during end-to-end testing)

student/src/test/java/tds/student/services/remote/RemotePrintServiceTest.java
burzum619: Trust me, I wanted to. Unfortunately the version of that library for Java 7 has a dependency on a newer version of the apache.commons.lang3 tools. I tried getting it to work briefly, but it looked like that dependency was being pulled in from a parent pom. I wasn't comfortable enough upgrading to the newer version for fear of introducing a regression, but if you guys want me to give it a shot and do some testing, we can. It has saved a huge amount of time on the exam side.

student.library/src/main/java/tds/student/services/abstractions/PrintService.java
burzum619: Two of these methods (the ones with the requestType parameter) have been converted to private methods - the public methods simply call this with the common method for page/passage.

student/src/main/java/tds/student/services/remote/RemoteAccommodationsService.java
burzum619: I don't think those are logical equivalents. We want to check if the number of accommodations assigned to a segment position is less than the number of segments + 1 (the 1 represents the "assessment" accommodations").

As a side note, this wasn't meant to be included in this PR as it was a bug fix I was briefly investigating and thought I stashed - it accidentally snuck its way in.

PR #31, opened 2017-03-09
Add tests to ensure backwards compatibility with existing code when m…

wtritch:
…aking future changes.

Minor code cleanup.

I'm going to leave this PR running during development for any outside suggestions/concerns.


SpellCheck/pom.xml
```diff
@@ -46,6 +46,24 @@
 			<artifactId>jna</artifactId>
 			<version>3.0.9</version>
 		</dependency>
+		<dependency>
+			<groupId>junit</groupId>
+			<artifactId>junit</artifactId>
+			<version>4.12</version>
```
millerts, line 52: Should we define these version in the parent?  I believe all of them can use these versions.

SpellCheck/src/main/java/AIR/SpellCheck/Engines/HunspellEngine.java
```diff
@@ -25,44 +27,52 @@
...
     try {
-      String languageCode = WebHelper.getQueryString ("lang");
-      if (_spellEngine.getDictionary (languageCode).misspelled(word)) {
-        suggestions.addAll (_spellEngine.getDictionary (languageCode).suggest(word));
-      }
-      if (_spellEngine.getDictionary (languageCode).misspelled(word)) {
-        suggestions.addAll (_spellEngine.getDictionary (languageCode).suggest(word));
+      if (spellEngineFactory.getDictionary (languageCode).misspelled(word)) {
+        suggestions.addAll (spellEngineFactory.getDictionary (languageCode).suggest(word));
       }
```
wtritch: I know this is a slight change in functionality.  Previously, the code was returning all the suggestions twice... but I just couldn't make myself do that... Any objections?

SpellCheck/src/main/java/AIR/SpellCheck/Engines/HunspellEngine.java
```diff
@@ -25,44 +27,52 @@
...
     boolean isCorrect = true;
     try {
-      String languageCode = WebHelper.getQueryString ("lang");
-      if (_spellEngine.getDictionary (languageCode).misspelled(word)) {
+      if (spellEngineFactory.getDictionary (languageCode).misspelled(word)) {
         isCorrect = false;
       }
-      if (_spellEngine.getDictionary (languageCode).misspelled(word)) {
-        isCorrect = false;
-      }
```
wtritch, line 42: Slight functionality change.  Previously, the code was checking if the word was misspelled twice.  ...and I have no idea why.  I changed it to only check once.

SpellCheck/src/main/java/AIR/SpellCheck/Engines/HunspellEngine.java
```diff
@@ -25,44 +27,52 @@
...
+    final String languageCode = getLanguage();
+    final List<String> suggestions = new ArrayList<> ();
     try {
-      String languageCode = WebHelper.getQueryString ("lang");
-      if (_spellEngine.getDictionary (languageCode).misspelled(word)) {
-        suggestions.addAll (_spellEngine.getDictionary (languageCode).suggest(word));
-      }
-      if (_spellEngine.getDictionary (languageCode).misspelled(word)) {
-        suggestions.addAll (_spellEngine.getDictionary (languageCode).suggest(word));
+      if (spellEngineFactory.getDictionary (languageCode).misspelled(word)) {
```
millerts: Can we remove the space since we're changing this block.

SpellCheck/src/main/java/AIR/SpellCheck/Engines/HunspellEngine.java
```diff
@@ -25,44 +27,52 @@
...
+        suggestions.addAll (spellEngineFactory.getDictionary (languageCode).suggest(word));
       }
-    } catch (Exception e) {
-      e.printStackTrace();
+    } catch (final Exception e) {
+      LOG.error("Unable to get spelling suggestions for word {} in language {}", word, languageCode);
```
millerts: Should we include the exception with the error?

SpellCheck/src/test/java/AIR/SpellCheck/Engines/HunspellEngineTest.java
```diff
@@ -0,0 +1,77 @@
...
+    private SpellEngineFactory spellEngineFactory;
+
+    @Mock
+    private Hunspell.Dictionary dictionary;
+
+    private HunspellEngine engine;
+
+    @Before
+    public void setup() {
+        engine = spy(new HunspellEngine(spellEngineFactory));
```
millerts, line 35: The old spy.  I almost had forgotten about that trick.

SpellCheck/src/test/java/AIR/SpellCheck/SpellEngineFactoryTest.java
```diff
@@ -0,0 +1,63 @@
+package AIR.SpellCheck;
+
+
```
millerts: extra line break

millerts (approved):
lgtm

SpellCheck/src/main/java/AIR/SpellCheck/Engines/HunspellEngine.java
wtritch: fixed

SpellCheck/src/main/java/AIR/SpellCheck/Engines/HunspellEngine.java
wtritch: doh, yes we should.  Added.

SpellCheck/src/test/java/AIR/SpellCheck/Engines/HunspellEngineTest.java
wtritch: Nice for wrapping and intercepting static method calls.  Still feels ugly, though.  The interface *should* have been designed with a language parameter rather than requiring a reach-back to manually parse a request param... But, one thing at a time.

SpellCheck/src/test/java/AIR/SpellCheck/SpellEngineFactoryTest.java
wtritch: fixed

SpellCheck/pom.xml
wtritch: We could... I'm going to leave that as a stretch task.  But it might help clean up some of our POMs.

PR #32, opened 2017-03-09
Replace embedded Hunspell with released Maven-Central dependency.

wtritch:
This PR removes our embedded Hunspell dependency and replaces it with a Java Hunspell bridge implementation available from maven central.


SpellCheck/pom.xml
```diff
@@ -9,14 +9,6 @@
...
 	</parent>
 
-	<repositories>
-	    <!--other repositories if any-->
-	    <repository>
-	        <id>project.local</id>
-	        <name>project</name>
-	        <url>file:${project.basedir}/repo</url>
-	    </repository>
-	</repositories>
```
wtritch, line 19: We don't need this any more.  As a benefit, we shouldn't see `${project.basedir}` directories being created on our filesystems any more.

SpellCheck/pom.xml
```diff
@@ -25,43 +17,41 @@
 			<comments>Educational Online Test Delivery System Copyright (c) 2013 American Institutes for Research</comments>
 		</license>
 	</licenses>
+
 	<dependencies>
+		<dependency>
+			<groupId>com.atlascopco</groupId>
+			<artifactId>hunspell-bridj</artifactId>
+			<version>1.0.4</version>
+		</dependency>
```
wtritch, line 26: This hunspell java bridge implementation is available from Maven Central. (Available under Mozilla Public License 1.1)

SpellCheck/pom.xml
```diff
@@ -25,43 +17,41 @@
...
-     	 <dependency>
-			<groupId>com.sun.jna</groupId>
-			<artifactId>jna</artifactId>
-			<version>3.0.9</version>
-		</dependency>
+
 		<dependency>
 			<groupId>junit</groupId>
 			<artifactId>junit</artifactId>
-			<version>4.12</version>
```
wtritch, line 52: I moved these version declarations into the parent pom and shared them among all the submodules in TDS_Student.

millerts (approved):

SpellCheck/src/main/java/AIR/SpellCheck/Engines/HunspellEngine.java
```diff
@@ -41,23 +42,21 @@ public boolean CheckWord(final String word) {
     final String languageCode = getLanguage();
     boolean isCorrect = true;
     try {
-      if (spellEngineFactory.getDictionary(languageCode).misspelled(word)) {
-        isCorrect = false;
-      }
+      isCorrect = spellEngineFactory.getDictionary(languageCode).isCorrect(word);
```
wtritch, line 45: Small API change in the underlying library here.  Rather than a negative ".misspelled" method, there is a positive "isCorrect" method.

PR #35, opened 2017-03-14
Feature/add client name for scoreable

millerts:


millerts:
This is ready to be reviewed.

student.library/src/main/java/tds/student/sql/data/ItemResponseScorable.java
```diff
@@ -32,22 +32,28 @@
...
-    _bankKey = bankKey;
-    _itemKey = itemKey;
-    _segmentID = segmentID;
-    _value = value;
-    _scoreMark = scoreMark;
-    _filePath = itemFilePath;
+    this(testKey, testID, language, position, sequence, bankKey, itemKey, segmentID, value, scoreMark, itemFilePath, "");
+  }
+
+  public ItemResponseScorable(String _testKey, String _testID, String _language, int _position, int _sequence, long _bankKey, long _itemKey, String _segmentID, String _value, UUID _scoreMark, String _filePath, String clientName) {
```
millerts: Exam scoring needs this change

student.library/src/main/java/tds/student/sql/data/OpportunityStatusExtensions.java
```diff
@@ -30,16 +28,6 @@ public static OpportunityStatusType parse (String statusString) {
     return parsedStatusType;
   }
 
-  public static OpportunityStatusType parseExamStatus(final String examStatusCode) {
```
millerts, line 33: This requires exam-client to be a dependency which then creates a circular dependency.  Instead I moved the mapper code to student proper.

student.library/src/main/java/tds/student/sql/data/ItemResponseScorable.java
jtreuting: Not a big deal but the other constructor and the clientName don't have underscores to start them.  Probably good to be consistent

student/src/main/java/tds/student/services/remote/ExamStatusMapper.java
```diff
@@ -0,0 +1,40 @@
...
+    examStatusToOpportunityStatus.put(ExamStatusCode.STATUS_SEGMENT_EXIT, OpportunityStatusType.SegmentExit);
+    examStatusToOpportunityStatus.put(ExamStatusCode.STATUS_CLOSED, OpportunityStatusType.Closed);
+    examStatusToOpportunityStatus.put(ExamStatusCode.STATUS_DISABLED, OpportunityStatusType.Disabled);
+  }
+
+  static OpportunityStatusType parseExamStatus(final String examStatusCode) {
+    OpportunityStatusType parsedStatusType = OpportunityStatusType.Unknown;
+
+    if (StringUtils.isNotEmpty(examStatusCode)) {
+      parsedStatusType = examStatusToOpportunityStatus.get (examStatusCode);
```
jtreuting, line 35: I know you copied this over, but seems like the check should be on `examStatusToOpportunityStatus.containsKey(examStatusCode)` otherwise if it doesn't exist is `parsedStatusType` null or does the call to `get` throw?

jtreuting (approved):
couple small notes

student/src/main/java/tds/student/services/remote/ExamStatusMapper.java
millerts: The logic in this flow isn't clear, but I was trying to mimic the implementation [here](https://github.com/SmarterApp/TDS_Student/blob/develop/student.library/src/main/java/tds/student/sql/data/OpportunityStatusExtensions.java#L22) which is what this replaces.  I'm fine with changing it to be more correct, but it will deviate from the legacy code.

student.library/src/main/java/tds/student/sql/data/OpportunityStatusExtensions.java
```diff
@@ -13,12 +13,10 @@
 import java.util.HashMap;
 import java.util.Map;
 
-import tds.exam.ExamStatusCode;
-
 public class OpportunityStatusExtensions
 {
   private static final Map<String, OpportunityStatusType> _statuses = new HashMap<> ();
-  private static final Map<String, OpportunityStatusType> examStatusToOpportunityStatus = new HashMap<>();
```
wtritch, line 21: These are static immutable lookup tables, not state?  So we could have two Student WARs running behind a load balancer?

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -164,15 +163,15 @@ public OpportunityStatus getStatus(final OpportunityInstance oppInstance) throws
...
+    status.setStatus(ExamStatusMapper.parseExamStatus(examApproval.getExamStatusCode()));
 
     return status;
   }
 
   @Override
   public boolean setStatus(final OpportunityInstance oppInstance, final OpportunityStatusChange statusChange) throws ReturnStatusException {
     boolean isApproved = false;
-    ReturnStatus returnStatus = null;
+    ReturnStatus returnStatus;
```
wtritch, line 174: I think we can declare this variable down where it is being set? Line 195?

student.library/src/main/java/tds/student/sql/data/OpportunityStatusExtensions.java
millerts: Yes.  These are just lookup tables that convert exam status code Strings to the existing legacy codes object types.

PR #46, opened 2017-03-24
Feature/event logging

trapnine:
New event logging to centralized ELK stack for performance analysis.


student/src/main/java/tds/student/tdslogger/StudentEventParserFactory.java
```diff
@@ -0,0 +1,48 @@
...
+  protected Map<String, Class<? extends EventParser>> getEventParserClassMap() {
+    return classMap;
+  }
+
+  public static class UpdateResponsesParser extends EventParser {
+    @Override
+    public Optional<EventInfo> parsePreHandle(final HttpServletRequest request, Object handler, EventLogger logger) {
+      String studentId = "TODO: fake Student ID enter";//request.getParameter("studentID");
+      final Map<EventLogger.EventData, Object> fields = getEventDataFields(request);
+      fields.put(StudentEventData.STUDENT_ID, studentId);
```
wtritch: Rather than getting and putting, should we just override the getEventDataFields(request) method?

student/src/main/java/tds/student/web/handlers/MasterShellHandler.java
```diff
@@ -136,6 +137,9 @@
   private ITDSLogger             _tdsLogger;
 
   @Autowired
+  private StudentEventLogger     _eventLogger;
```
wtritch, line 140: If we autowire a request-scoped property, does that mean this class needs to also be request-scoped?

student/src/main/resources/root-context.xml
```diff
@@ -136,6 +137,8 @@
 	<beans:bean id="restTemplate" class="org.springframework.web.client.RestTemplate">
 		<beans:constructor-arg ref="clientRequestFactory"/>
 	</beans:bean>
+	<beans:bean id="objectMapper" class="com.fasterxml.jackson.databind.ObjectMapper" />
```
wtritch: Student doesn't already have an injectible ObjectMapper bean?

student/src/main/java/tds/student/web/handlers/TestResponseHandler.java
```diff
@@ -141,8 +166,17 @@ private void updateResponses (HttpServletRequest request, HttpServletResponse re
...
+      _eventLogger.info(((EventInfo)request.getAttribute(LoggingInterceptor.EVENT_INFO))
+        .withMessage("_itemScoringService.updateResponses")
+        .withCheckpoint(SERVICE_CALL.name()));
+
       List<ItemResponseUpdateStatus> responseResults = null;
         responseResults = _itemScoringService.updateResponses (testOpp.getOppInstance (), responseReader.getResponses (), responseReader.getPageDuration());
+
+      _eventLogger.info(((EventInfo)request.getAttribute(LoggingInterceptor.EVENT_INFO))
+        .withMessage("_itemScoringService.updateResponses")
+        .withCheckpoint(SERVICE_RETURN.name()));
```
wtritch: I hate to say it, but this still feels pretty invasive at the @Controller level... What I'd love is to just be able to call:
```
LOG.info("updating responses");
```
using the standard SLF4J logger down inside the ItemScoringService implementation and have it be bound to the overall request log set.

student/src/main/resources/root-context.xml
trapnine: removed and updated.

student/src/main/java/tds/student/web/handlers/TestResponseHandler.java
trapnine: We have a separate task to log in ItemScoringService - I'll remove this gratuitous entry for now.

student/src/main/java/tds/student/tdslogger/StudentEventParserFactory.java
trapnine: This was a sample parser and has been removed. Apologies.

student/src/main/java/tds/student/web/handlers/MasterShellHandler.java
trapnine: The StudentEventLogger class is annotated with a magic Spring 'scoped proxy mode' which arranges for a new instance to be created for each request (I tested it by logging from the constructor, and indeed a new instance is made for every single request).

That's in StudentEventLogger.java:12:
`@Scope(value = "request", proxyMode = ScopedProxyMode.TARGET_CLASS)`

student/pom.xml
```diff
@@ -24,7 +24,7 @@
 	<properties>
 		<progman-client.version>3.0.1</progman-client.version>
 		<sb11-mna-client.version>3.0.0</sb11-mna-client.version>
-		<sb11-shared-code.version>3.0.0</sb11-shared-code.version>
+		<sb11-shared-code.version>3.0.1-SNAPSHOT</sb11-shared-code.version>
```
millerts: Looks like we should release the shared code and then bump these versions to release.  you mentioned it on @mtrupkin's review as well.  I can help you do this or you can follow the steps [here](https://confluence.fairwaytech.com/display/SBAC1719/Releasing)

snapshot dependency references should no longer be committed to develop.

student/src/main/java/tds/student/tdslogger/StudentEventLogger.java
```diff
@@ -0,0 +1,36 @@
...
+
+  public enum StudentLogEvent implements LogEvent {
+    DETERMINE_NEXT_QUESTION,
+  }
+
+  public enum StudentEventData implements EventData {
+    STUDENT_ID,
+  }
+
+  public StudentEventLogger() {
```
millerts, line 37: Why have this constuctor?  If this is used you will have the default object mapper which differs from the one the application uses.

student/src/main/java/tds/student/web/handlers/TestResponseHandler.java
```diff
@@ -128,6 +137,17 @@ private void updateResponses (HttpServletRequest request, HttpServletResponse re
       // get the request information from tehe browser
       TestResponseReader responseReader = TestResponseReaderSax.parseSax (request.getInputStream (), testOpp);
 
+      List<Map<String,String>> responses = new ArrayList<>();
+      for(ItemResponseUpdate responseUpdate: responseReader.getResponses()) {
```
millerts: I mentioned this is in @mtrupkin 's review, but I wonder if there is any value in applications being able to register `ObjectLogEventHandlers` that could handle these types of event logging.  So the block from 140 - 149 would be in one of the handlers and this would be reduced to `_eventLogger.info(responseReader)` with the event logger knowing how to get the data it wants rather than sprinkling that logic into the controllers

millerts: That comment would be for any time you're populating data structures with bits of data from other objects.

student/src/main/java/tds/student/tdslogger/StudentEventLogger.java
trapnine: I followed a boilerplate with this from proctor. I asked @mtrupkin the same question. I tried removing it and it fails to start up without it. Is a no-arg constructor required? I noted while testing construction that one instance seems to be created before a request comes in. I can see which one it's calling on that first construction.

trapnine: I've confirmed that the 0-arg is called during startup, and the 1-arg is called for each request.

millerts: I wonder in this rare case if it would be better to Autowire the field so the correct ObjectMapper is injected since something in the application is calling this directly at startup?

millerts (approved):

PR #48, opened 2017-03-28
Adding browser user agent to start exam call

burzum619:


student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -268,9 +272,12 @@ public TestConfig startTest(final OpportunityInstance oppInstance, final String
     if (!isRemoteExamCallsEnabled) {
       return testConfig;
     }
-  
+
+    final HttpServletRequest httpRequest = HttpContext.getCurrentContext().getRequest();
```
millerts: Can we do this at the controller layer and modify the `OpportunityInstance` to have this value?  HttpRequest is specific to the endpoint.  This will blow up if this service would ever be called from a non Http flow, correct?

millerts: I'm actually not sure if [it'll throw](https://github.com/SmarterApp/SS_SharedMultiJar/blob/82ba1490747dda2f64ae2a1343f5089431d506fa/shared-web/src/main/java/AIR/Common/Web/Session/HttpContext.java) but I still feel this is a concern for the controller layer not the service layer.

millerts (changes requested):

PR #51, opened 2017-04-06
Fix health checks.

wtritch:
Remove health checks for services that will no longer be directly called by Student.

This PR removes health checks on:
1. the python equation-scoring webapp
2. the StudentReportProcessor
Neither of these services will be directly called by Student.

This PR also corrects the health check on the TDS_ExamService. We were making calls to
http://exam-host:port/exam/health
and getting a 404 rather than against
http://exam-host:port/health

### DIFF
--- a/student/src/main/java/tds/student/diagnostic/services/impl/StudentDiagnosticDependencyServiceImpl.java
+++ b/student/src/main/java/tds/student/diagnostic/services/impl/StudentDiagnosticDependencyServiceImpl.java
@@ -60,37 +60,14 @@ public class StudentDiagnosticDependencyServiceImpl extends AbstractDiagnosticDe
 
         statusList.add(getArt());
         statusList.add(getProgman());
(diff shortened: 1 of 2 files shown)


student/src/main/java/tds/student/diagnostic/services/impl/StudentDiagnosticDependencyServiceImpl.java
```diff
@@ -60,38 +60,15 @@ public Providers getProviders() {
...
 
         return new Providers(statusList);
     }
 
-    protected List<String> getProcesses(String grepValue) {
-        List<String> processes = new ArrayList<>();
-
-        try {
-            String line;
-            Process p = Runtime.getRuntime().exec("ps -ax"); // for some reason | grep is not working as expected
```
wtritch, line 75: This was a fun little discovery.

millerts (approved):

PR #59, opened 2017-04-18
Log student app startup and shutdown.

trapnine:


mtrupkin (approved):
StudentEventLogger could be final.

student/src/main/java/tds/student/tdslogger/ApplicationShutdownLogger.java
```diff
@@ -0,0 +1,31 @@
...
+
+  @Autowired
+  public ApplicationShutdownLogger(@Qualifier("integrationObjectMapper") final ObjectMapper objectMapper) {
+    logger = new StudentEventLogger(objectMapper);
+  }
+
+  @Override
+  public void onApplicationEvent(final ContextClosedEvent event) {
+    // This event fires multiple times - we only want to log the first occurrence.
+    if (!logged) {
```
millerts: This class is a singleton (by default all spring beans are) so this will only be logged once for an entire applications lifecycle.  To get it to log again it'll require a restart of the application server

Not sure if that is the intention.  If that is the intention I would appreciate a little more robust comments in here explaining what this is doing and why it only logs once.  For future maintainers that do not know anything about the logging work.

student/src/main/java/tds/student/tdslogger/ApplicationStartupLogger.java
```diff
@@ -0,0 +1,31 @@
...
+
+  @Autowired
+  public ApplicationStartupLogger(@Qualifier("integrationObjectMapper") final ObjectMapper objectMapper) {
+    logger = new StudentEventLogger(objectMapper);
+  }
+
+  @Override
+  public void onApplicationEvent(final ContextRefreshedEvent event) {
+    // This event fires multiple times - we only want to log the first occurrence.
+    if (!logged) {
```
millerts: Same comment as in the other class.

millerts (commented):
I would like to see more documentation about the purpose of these classes in general and then more detail on why we only log once.

student/src/main/java/tds/student/tdslogger/ApplicationShutdownLogger.java
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.context.ApplicationListener;
+import org.springframework.context.event.ContextClosedEvent;
+import org.springframework.stereotype.Component;
+
+import static tds.student.tdslogger.StudentEventLogger.StudentLogEvent.APP_SHUTDOWN;
+
+@Component
+public class ApplicationShutdownLogger implements ApplicationListener<ContextClosedEvent> {
```
wtritch, line 13: I'd be tempted to combine these two classes into a single ApplicationEventLogger

student/src/main/java/tds/student/tdslogger/ApplicationStartupLogger.java
```diff
@@ -0,0 +1,31 @@
...
+  private StudentEventLogger logger;
+  boolean logged = false;
+
+  @Autowired
+  public ApplicationStartupLogger(@Qualifier("integrationObjectMapper") final ObjectMapper objectMapper) {
+    logger = new StudentEventLogger(objectMapper);
+  }
+
+  @Override
+  public void onApplicationEvent(final ContextRefreshedEvent event) {
```
wtritch: @trapnine I'm a little rusty on my application events, but will this get fired when the configuration service pushes  a configuration update event?

millerts (approved):
I think a single class would be a bit better, but it isn't a big deal to me.

If we could add a bit of documentation on the "why" logged once that'd be nice.

student/src/main/java/tds/student/tdslogger/ApplicationShutdownLogger.java
trapnine: I tried to do that initially, but couldn't get it to work due to having to effectively implement the same interface twice to make the filtering work (the applied-to generic types get type-deleted).

student/src/main/java/tds/student/tdslogger/ApplicationStartupLogger.java
trapnine: Yes. It is fired when properties, xml or any schema files are loaded/refreshed. I'm using it to log the initial load at startup, which is a common approach in spring 3.

PR #62, opened 2017-04-25
Respect X-FORWARDED headers when making a self-call to generate...

wtritch:
dialog content.

When populated a tutorial dialog, TDS_Student forges a new HTTP request back into itself to use ApacheFaces to generate the dialog content through a MockHttpRequest.
This fixes a problem when TDS_Student is hosted behind a load balancer, where the self-request cannot be generated as a copy of the original HttpServletRequest.  The server must respect the X-FORWARDED headers to be able to find itself.

### DIFF
--- a/student/src/main/java/tds/student/web/backing/DialogFrameBacking.java
+++ b/student/src/main/java/tds/student/web/backing/DialogFrameBacking.java
@@ -23,12 +23,15 @@ import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
 import org.springframework.http.ResponseEntity;
 
+import org.springframework.web.util.UriComponentsBuilder;
 import tds.student.web.StudentContext;
 import tds.student.web.handlers.DialogFrameHandler;
 import AIR.Common.Web.WebHelper;
 import AIR.Common.Web.Session.HttpContext;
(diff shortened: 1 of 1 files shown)


student/src/main/java/tds/student/web/backing/DialogFrameBacking.java
```diff
@@ -67,18 +74,29 @@ public void setOnLoad ()
...
+
+      if (isNotBlank(request.getHeader(X_FORWARDED_HOST))) {
+          uriBuilder.host(request.getHeader(X_FORWARDED_HOST));
+      }
+      if (isNotBlank(request.getHeader(X_FORWARDED_PORT))) {
+          uriBuilder.port(Integer.valueOf(request.getHeader(X_FORWARDED_PORT), 10));
+      }
+      if (isNotBlank(request.getHeader(X_FORWARDED_PROTOCOL))) {
+          uriBuilder.scheme(request.getHeader(X_FORWARDED_PROTOCOL));
+      }
```
wtritch, line 91: I worry that there are other places in the legacy code that are also creating new HTTP requests to themselves, but I didn't find any other usages of the GenericRestAPIClient class.

millerts (approved):

PR #63, opened 2017-04-25
Add remote item scoring service implementation.

wtritch:
This PR won't be merged until after the scoring implementation in TDS_ExamService is complete and tested, but I wanted to put the PR to speed up the eventual commit.

### DIFF
--- a/student/src/main/java/tds/student/sbacossmerge/data/TestResponseReaderSax.java
+++ b/student/src/main/java/tds/student/sbacossmerge/data/TestResponseReaderSax.java
@@ -117,6 +117,7 @@ public class TestResponseReaderSax extends TestResponseReader
         _itemResponseUpdate.setTestKey (_testOpp.getTestKey ());
         _itemResponseUpdate.setLanguage (_testOpp.getLanguage ());
         _itemResponseUpdate.setTestID (_testOpp.getTestID ());
+        _itemResponseUpdate.setClientName(_testOpp.getOppInstance().getExamClientName());
 
         getResponses ().add (_itemResponseUpdate);
         popHandler ();
--- a/student/src/main/java/tds/student/services/ItemScoringService.java
+++ b/student/src/main/java/tds/student/services/ItemScoringService.java
@@ -39,6 +39,7 @@ import AIR.Common.Web.WebValueCollection;
 import AIR.Common.Web.WebValueCollectionCorrect;
 import TDS.Shared.Data.ReturnStatus;
 import TDS.Shared.Exceptions.ReturnStatusException;
+import org.springframework.stereotype.Service;
 import tds.itemrenderer.data.AccLookup;
 import tds.itemrenderer.data.IITSDocument;
 import tds.itemrenderer.data.ITSMachineRubric;
@@ -69,7 +70,7 @@ import tds.student.sql.data.OpportunityInstance;
 import tds.student.sql.singletons.ClientSingleton;
 import tds.student.tdslogger.TDSLogger;
 
-@Component
+@Service("legacyItemScoringService")
 @Scope ("prototype")
 public class ItemScoringService implements IItemScoringService
 {
--- a/student/src/main/java/tds/student/services/remote/RemoteAdaptiveService.java
+++ b/student/src/main/java/tds/student/services/remote/RemoteAdaptiveService.java
@@ -51,14 +51,14 @@ public class RemoteAdaptiveService implements IAdaptiveService {
       return pageGroup;
     }
 
-    List<OpportunityItem> items = examItemResponseRepository.createNextItemGroup(testOpportunity.getOppInstance().getExamId(), lastPage, lastPosition);
+    final List<OpportunityItem> items = examItemResponseRepository.createNextItemGroup(testOpportunity.getOppInstance().getExamId(), lastPage, lastPosition);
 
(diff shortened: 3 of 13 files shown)


student/src/main/java/tds/student/sbacossmerge/data/TestResponseReaderSax.java
```diff
@@ -117,6 +117,7 @@ public void endElement (String uri, String localName, String qName) throws SAXEx
         _itemResponseUpdate.setTestKey (_testOpp.getTestKey ());
         _itemResponseUpdate.setLanguage (_testOpp.getLanguage ());
         _itemResponseUpdate.setTestID (_testOpp.getTestID ());
+        _itemResponseUpdate.setClientName(_testOpp.getOppInstance().getExamClientName());
```
wtritch, line 120: I needed this field to be populated on the ItemResponseUpdate objects.

student/src/main/java/tds/student/services/ItemScoringService.java
```diff
@@ -69,7 +70,7 @@
 import tds.student.sql.singletons.ClientSingleton;
 import tds.student.tdslogger.TDSLogger;
 
-@Component
+@Service("legacyItemScoringService")
```
wtritch, line 73: We don't have to annotate classes with both \@Service and \@Component, Spring will register either one for DI.

student/src/main/java/tds/student/services/remote/RemoteItemScoringService.java
```diff
@@ -0,0 +1,108 @@
...
+
+    @Override
+    public ItemScore checkScoreability(final IItemResponseScorable responseScorable, final IITSDocument itsDoc) throws ReturnStatusException {
+        if (legacyCallsEnabled) {
+            return legacyItemScoringService.checkScoreability(responseScorable, itsDoc);
+        }
+
+        //This method is only called via a legacy service self-reference and via the unused TimerScoringService
+        throw new UnsupportedOperationException("No remote implementation of checkScoreability");
+    }
```
wtritch, line 85: Some of these methods don't have a remote implementation because they're only called from inside the legacy implementation OR unused "Timer" classes.

student/src/main/java/tds/student/sql/repository/remote/impl/RemoteItemScoringRepository.java
```diff
@@ -0,0 +1,76 @@
...
+            return restTemplate.exchange(
+                builder.build().toUri(),
+                HttpMethod.PUT,
+                requestHttpEntity,
+                new ParameterizedTypeReference<List<ItemResponseUpdateStatus>>() {
+                }).getBody();
+        } catch (final HttpStatusCodeException e) {
+            final ReturnStatusException statusException = new ReturnStatusException("Failed to update response: " + e.getResponseBodyAsString());
+            statusException.getReturnStatus().setHttpStatusCode(500);
+            throw statusException;
```
wtritch, line 71: I wanted to see the advertised reason of the failure in the exception message.

millerts: Good call.  I'll add that to my latest PR as well.

student/src/test/java/tds/student/services/remote/RemoteItemScoringServiceTest.java
```diff
@@ -0,0 +1,142 @@
...
+    public void itShouldUseRepositoryAndLegacyToUpdateResponsesWhenEnabled() throws Exception {
+        final OpportunityInstance oppInstance = oppInstance();
+        final List<ItemResponseUpdate> responsesUpdated = newArrayList(mock(ItemResponseUpdate.class));
+        final Float pageDuration = 1.23f;
+
+        service.updateResponses(oppInstance, responsesUpdated, pageDuration);
+
+        verify(mockLegacyItemScoringService).updateResponses(eq(oppInstance), eq(responsesUpdated), eq(pageDuration));
+        verify(mockRepository).updateResponses(
+            eq(oppInstance.getExamId()),
```
millerts: Do you need the `eq`?  I believe it will work without it.

millerts (approved):
LGTM

student/src/test/java/tds/student/services/remote/RemoteItemScoringServiceTest.java
wtritch: good point, since they're all equality matchers, I could just use the values.  Habit.

student/src/main/java/tds/student/services/remote/RemoteAdaptiveService.java
```diff
@@ -51,14 +51,14 @@ public PageGroup createNextItemGroup(final TestOpportunity testOpportunity, fina
...
 
-    for (OpportunityItem item : items) {
-      ItemResponse response = new ItemResponse(item);
-      response.setPrefetched(true);
-      remotePageGroup.add(response);
+    if (remotePageGroup != null) {
+      for (final ItemResponse response : remotePageGroup) {
+        response.setPrefetched(true);
+      }
     }
```
wtritch, line 62: Added this to handle the end of a test, when there are no additional item groups to be created.  I also changed PageGroup to use the static Create factory method.  (Note that passing the factory an empty collection results in a NULL PageGroup value.)

millerts (approved):

PR #64, opened 2017-04-25
Feature/remote item selection

millerts:


pom.xml
```diff
@@ -22,7 +22,7 @@
         <mockito.version>1.10.19</mockito.version>
         <tds-dll.version>3.0.0</tds-dll.version>
         <test-scoring.version>3.0.0</test-scoring.version>
-        <tds-student-common.version>3.0.3</tds-student-common.version>
+        <tds-student-common.version>3.0.5-SNAPSHOT</tds-student-common.version>
```
millerts: I will clean up this file and point it to the latest versions.

student/pom.xml
```diff
@@ -25,7 +25,7 @@
 		<progman-client.version>3.0.1</progman-client.version>
 		<sb11-mna-client.version>3.0.0</sb11-mna-client.version>
 		<sb11-shared-code.version>3.0.4</sb11-shared-code.version>
-		<tds-exam-client.version>0.0.12</tds-exam-client.version>
+		<tds-exam-client.version>0.0.13-SNAPSHOT</tds-exam-client.version>
```
millerts: I will clean up this file and point to latest versions once things are released and before merging.

student/src/main/java/tds/student/services/remote/RemoteAdaptiveService.java
```diff
@@ -0,0 +1,66 @@
...
+
+    PageGroup remotePageGroup = new PageGroup(items.get(0));
+
+    for (OpportunityItem item : items) {
+      ItemResponse response = new ItemResponse(item);
+      response.setPrefetched(true);
+      remotePageGroup.add(response);
+    }
+
+    return remotePageGroup;
```
millerts: I'm going to switch this to always return the page group created by the legacy code for now.  That way QA can verify the data is correct.  The PageGroup returned from this call is then used to insert items into the legacy database.

I could add assertions in here, which i had in before, but I think it is a bit overkill.  

@jtreuting - This is where the legacy and remote calls starts to break down.

student/src/test/java/tds/student/sql/repository/remote/RemoteExamItemResponseRepositoryTest.java
```diff
@@ -38,7 +49,79 @@ public void shouldMarkForReviewSuccessful() throws ReturnStatusException {
...
+
+        URI uri = UriComponentsBuilder.fromHttpUrl(String.format("http://localhost:8080/exam/%s/item/%d/review",
+          opportunityInstance.getExamId(), 3))
+          .queryParam("sessionId", opportunityInstance.getSessionKey())
+          .queryParam("browserId", opportunityInstance.getExamBrowserKey())
+          .build()
+          .toUri();
+
         remoteExamItemResponseRepository.markItemForReview(opportunityInstance, position, mark);
-        verify(mockRestTemplate).exchange(isA(URI.class), isA(HttpMethod.class), isA(HttpEntity.class), isA(ParameterizedTypeReference.class));
```
millerts, line 42: This test was just testing that the restTemplate was called.  Fixed it to actually verify the call has the right data.

student/src/main/java/tds/student/services/remote/RemoteAdaptiveService.java
wtritch: Maybe just a `return legacyCallsEnabled ? pageGroup : remotePageGroup;` so we don't need a code change when we disable legacy?

wtritch (approved):
LGTM once the Pom is clean

pom.xml
```diff
@@ -124,7 +124,7 @@
                 <groupId>org.assertj</groupId>
                 <artifactId>assertj-core</artifactId>
                 <version>${assertj.version}</version>
-                <scope>test</scope>
+                <!--<scope>test</scope>-->
```
wtritch: This is going to be changed back to a test-only dependency?

student/pom.xml
```diff
@@ -330,6 +336,10 @@
             <artifactId>jsp-api</artifactId>
             <scope>provided</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.assertj</groupId>
+			<artifactId>assertj-core</artifactId>
```
wtritch, line 341: test-scoped dependency?

millerts: Fixed all the assertj test scopes.  Nice catch.  I was using them to assert the values were the same in the remote and legacy objects.

PR #66, opened 2017-04-28
Added null guards to cookie parsing

hbgwhite:
Added null guards to cookie parsing

### DIFF
--- a/student/src/main/java/tds/student/web/StudentContext.java
+++ b/student/src/main/java/tds/student/web/StudentContext.java
@@ -164,12 +164,19 @@ public class StudentContext
     if (oppInfo.getBrowserKey() != null) {
       StudentCookie.setCookieData ("O_BKEY", oppInfo.getBrowserKey ().toString ());
     }
-
-    TDSIdentity.getCurrentTDSIdentity().setAuthCookieValue(EXAM_CLIENT_NAME, oppInfo.getExamClientName());
-    TDSIdentity.getCurrentTDSIdentity().setAuthCookieValue(EXAM_ID_COOKIE_KEY, oppInfo.getExamId().toString());
+    if(oppInfo.getExamClientName() != null) {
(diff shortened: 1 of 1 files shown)


jtreuting (commented):
Will the app actually work if that data is not provided?  @hbgwhite Why do you believe this is needed?  I'm guessing it's for the new spanish tests, but I think we need to error on the side of making the packages work with the current system than altering the current system, unless there is no other way.

hbgwhite:
@jtreuting - There was a bug in the code where it would attempt to parse out unnecessary Exam information when the switches that used that information were disabled in the Progman profile. This information is not present when running in legacy mode, and was causing exceptions when the application attempted to retrieve it.

student/src/main/java/tds/student/web/StudentContext.java
```diff
@@ -164,12 +164,19 @@ public static void saveOppInfo (String testKey, String testID, OpportunityInfo o
     if (oppInfo.getBrowserKey() != null) {
       StudentCookie.setCookieData ("O_BKEY", oppInfo.getBrowserKey ().toString ());
     }
-
-    TDSIdentity.getCurrentTDSIdentity().setAuthCookieValue(EXAM_CLIENT_NAME, oppInfo.getExamClientName());
-    TDSIdentity.getCurrentTDSIdentity().setAuthCookieValue(EXAM_ID_COOKIE_KEY, oppInfo.getExamId().toString());
+    if(oppInfo.getExamClientName() != null) {
```
millerts, line 167: @jtreuting, @hbgwhite and I discussed this change.  Right now this blows up if you are running with the `remoteEnabled` flag turned off.  Other areas of the application will throw if remote is on and these aren't populated which I think is fine.  We can probably remove the null checks once we turn off the legacy for good.

millerts (approved):

PR #67, opened 2017-04-28
Integration ExamAccommodationMetadata call

burzum619:
Integrating ExamAccommodationMetadata call to student

### DIFF
--- a/student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
+++ b/student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
@@ -1,9 +1,13 @@
 package tds.student.services.remote;
 
+import TDS.Shared.Browser.BrowserAction;
 import TDS.Shared.Browser.BrowserInfo;
+import TDS.Shared.Browser.BrowserRule;
+import TDS.Shared.Browser.BrowserValidation;
 import TDS.Shared.Data.ReturnStatus;
 import TDS.Shared.Exceptions.ReturnStatusException;
 import com.google.common.base.Optional;
+import org.apache.commons.lang3.StringUtils;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -12,6 +16,7 @@ import org.springframework.beans.factory.annotation.Value;
 import org.springframework.context.annotation.Scope;
 import org.springframework.stereotype.Service;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.UUID;
 
@@ -19,6 +24,7 @@ import tds.common.Response;
 import tds.common.ValidationError;
 import tds.exam.Exam;
 import tds.exam.ExamApproval;
+import tds.exam.ExamAssessmentMetadata;
 import tds.exam.ExamConfiguration;
 import tds.exam.ExamSegment;
 import tds.exam.ExamStatusCode;
@@ -38,6 +44,7 @@ import tds.student.sql.data.TestSegment;
 import tds.student.sql.data.TestSelection;
 import tds.student.sql.data.TestSession;
 import tds.student.sql.data.Testee;
+import tds.student.sql.repository.ConfigRepository;
 import tds.student.sql.repository.remote.ExamRepository;
 import tds.student.sql.repository.remote.ExamSegmentRepository;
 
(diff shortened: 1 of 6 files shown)


student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -380,4 +402,70 @@ private static TestConfig mapExamConfigurationToTestConfig(ExamConfiguration exa
...
+      selection.setTestID(metadata.getAssessmentId());
+      selection.setOpportunity(metadata.getAttempt());
+      selection.setMode("online");
+      selection.setDisplayName(metadata.getAssessmentLabel());
+      selection.setMaxOpportunities(metadata.getMaxAttempts());
+      selection.setSortOrder(0); // not used by UI
+      selection.setSubject(metadata.getSubject());
+      selection.setGrade(metadata.getGrade());
+      //TODO: Determine if we need this
+//      selection.getRequirements ().lookup (getTdsSettings ().getClientName (), testSelection);
```
wtritch: I'm not sure what our process is, do we allow merges with new TODOs?

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -380,4 +402,70 @@ private static TestConfig mapExamConfigurationToTestConfig(ExamConfiguration exa
...
+      validateBrowserInfo(browserInfo, selection);
+
+      selections.add(selection);
+    }
+
+    return selections;
+  }
+
+  private void validateBrowserInfo(final BrowserInfo browserInfo, final TestSelection selection) throws ReturnStatusException {
+    if (browserInfo != null) {
```
wtritch: Recommend switching this to guard code to reduce nesting. (Same for the nested if's below)

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
```diff
@@ -380,4 +402,70 @@ private static TestConfig mapExamConfigurationToTestConfig(ExamConfiguration exa
...
+      if (selection.getTestStatus() != TestSelection.Status.Disabled && selection.getTestStatus() != TestSelection.Status.Hidden) {
+        BrowserValidation browserValidation = new BrowserValidation();
+
+        for (BrowserRule rule : configRepository.getBrowserTestRules(selection.getTestID())) {
```
wtritch: Is there any way to pull this repository call out of the loop on line 410 and get the browser rules for all TestSelections in a single call?

student/src/main/java/tds/student/sql/repository/remote/impl/RemoteExamRepository.java
```diff
@@ -317,4 +317,31 @@ public void waitForSegmentApproval(final UUID examId, final SegmentApprovalReque
...
+
+    try {
+      responseEntity = restTemplate.exchange(
+          builder.build().encode().toUri(),
+          HttpMethod.GET,
+          requestHttpEntity,
+          new ParameterizedTypeReference<Response<List<ExamAssessmentMetadata>>>() {
+          });
+    } catch (RestClientException rce) {
+      throw new ReturnStatusException(rce);
```
wtritch: I like to see the advertised failure message in the thrown exception.  Maybe something like:
```
        } catch (final HttpStatusCodeException e) {
            final ReturnStatusException statusException = new ReturnStatusException("Failed to find assessment info: " + e.getResponseBodyAsString());
            statusException.getReturnStatus().setHttpStatusCode(500);
            throw statusException;
        } catch (final RestClientException rce) {
            throw new ReturnStatusException(rce);
        }
```
?

wtritch (approved):
LGTM

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
burzum619: I'll remove the todo/commented out code - it does not appear that we need to make this call.

student/src/main/java/tds/student/services/remote/RemoteOpportunityService.java
burzum619: Not without updating their ConfigRepository code, or creating a new endpoint in our ConfigService. Could do that, but for pragmatic purposes, and also because I don't think this is called often in practice, I that might be a bit of a premature optimization at the moment.

student/src/main/java/tds/student/sql/repository/remote/impl/RemoteExamRepository.java
millerts: We haven't been doing this but after dealing with all the issues in item selection I agree it is a good idea

millerts (approved):
lgtm

PR #69, opened 2017-04-30
Modified the interface to take in an OpportunityInstance so we can have

millerts:
Implementation details fior allowing `isTestComplete` to function for both remote and legacy. Interface change [here](https://github.com/SmarterApp/TDS_StudentCommon/pull/16)

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -22,7 +22,7 @@
         <mockito.version>1.10.19</mockito.version>
         <tds-dll.version>3.0.0</tds-dll.version>
         <test-scoring.version>3.0.0</test-scoring.version>
-        <tds-student-common.version>3.0.5</tds-student-common.version>
+        <tds-student-common.version>3.0.6-SNAPSHOT</tds-student-common.version>
     </properties>
 
     <modules>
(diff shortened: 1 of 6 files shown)


pom.xml
```diff
@@ -22,7 +22,7 @@
         <mockito.version>1.10.19</mockito.version>
         <tds-dll.version>3.0.0</tds-dll.version>
         <test-scoring.version>3.0.0</test-scoring.version>
-        <tds-student-common.version>3.0.5</tds-student-common.version>
+        <tds-student-common.version>3.0.6-SNAPSHOT</tds-student-common.version>
```
millerts: I will be bumping this to the latest version shortly

student/src/test/java/tds/student/services/ResponseServiceTest.java
```diff
@@ -204,44 +204,6 @@ public void testGetItemGroupFailure () throws SQLException, ReturnStatusExceptio
   // Sucess Test Case
   @Test
   @Ignore("Never actually implemented")
-  public void testisTestComplete () throws SQLException, ReturnStatusException {
```
millerts, line 207: Chose to just delete these tests rather than make them work since they were already ignored and code that we will hopefully remove in the 2018 release.

jtreuting (approved):

PR #70, opened 2017-05-01
TDS-519: Feature/review exam

jeffjohnson9046:
[TDS-519](https://jira.fairwaytech.com/browse/TDS-519):  Integration with exam service endpoint for review exam

**NOTE:** This PR depends on [this PR against the **TDS_ExamService**](https://github.com/SmarterApp/TDS_ExamService/pull/170)

**NOTE:** I Could not find a good way to unit test the `ReviewTestService` that was introduced as part of this PR.  Part of the issue is the reliance on the `TestManager` class and the way it gets at its dependencies (using `AIR.Common.Utilities.SpringApplicationContext.getBean()`.  I tried to test the "exam service" side, but also had some wiring issues for dependencies.  In the end, testing was conducted manually which is... disappointing.  If there are any ideas on how to write unit tests for the `ReviewTestService`, I'm certainly willing to listen.

### DIFF
--- a/student/pom.xml
+++ b/student/pom.xml
@@ -24,7 +24,7 @@
 	<properties>
 		<progman-client.version>3.0.1</progman-client.version>
 		<sb11-mna-client.version>3.0.0</sb11-mna-client.version>
-		<tds-exam-client.version>0.0.14</tds-exam-client.version>
+		<tds-exam-client.version>0.0.15-SNAPSHOT</tds-exam-client.version>
 		<sb11-shared-code.version>3.0.5</sb11-shared-code.version>
 		<tds-common-legacy.version>0.0.6</tds-common-legacy.version>
 		<tds-assessment-client.version>0.0.13</tds-assessment-client.version>
--- a/student/src/main/java/tds/student/services/ReviewTestServiceImpl.java
+++ b/student/src/main/java/tds/student/services/ReviewTestServiceImpl.java
@@ -64,7 +64,7 @@ public class ReviewTestServiceImpl implements ReviewTestService {
             return responseData;
         }
 
-        Optional<ValidationError> maybeValidationError = examRepository.reviewExam(testOpportunity.getOppInstance());
+        final Optional<ValidationError> maybeValidationError = examRepository.reviewExam(testOpportunity.getOppInstance());
         if (maybeValidationError.isPresent()) {
             tdsLogger.applicationError (maybeValidationError.get().getMessage(), "reviewTest", request, null);
             HttpContext.getCurrentContext()
@@ -81,8 +81,9 @@ public class ReviewTestServiceImpl implements ReviewTestService {
      * Executes the legacy implementation of review exam logic.
      *
      * @param testOpportunity The {@link tds.student.services.data.TestOpportunity}
-     * @param request
-     * @return
+     * @param request The HTTP request from the student application for reviewing the
+     *        {@link tds.student.services.data.TestOpportunity}
+     * @return A {@link AIR.Common.data.ResponseData} indicating success or failure
      * @throws ReturnStatusException
      * @throws IOException
      */
@@ -120,7 +121,7 @@ public class ReviewTestServiceImpl implements ReviewTestService {
 
         // put test in review mode
         OpportunityStatusChange statusChange = new OpportunityStatusChange (OpportunityStatusType.Review, true);
-        opportunityService.setStatus (testOpportunity.getOppInstance (), statusChange);
+        opportunityService.setStatus(testOpportunity.getOppInstance(), statusChange);
 
         return new ResponseData<>(TDSReplyCode.OK.getCode(), "OK", null);
     }
--- a/student/src/main/java/tds/student/services/ReviewTestServiceImpl.java
+++ b/student/src/main/java/tds/student/services/ReviewTestServiceImpl.java
@@ -1,0 +1,127 @@
+package tds.student.services;
+
+import AIR.Common.TDSLogger.ITDSLogger;
+import AIR.Common.Web.Session.HttpContext;
+import AIR.Common.Web.TDSReplyCode;
+import AIR.Common.data.ResponseData;
+import TDS.Shared.Exceptions.ReturnStatusException;
+import com.google.common.base.Optional;
+import org.apache.http.HttpStatus;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
+import org.springframework.beans.factory.annotation.Value;
+import org.springframework.context.annotation.Scope;
+import org.springframework.stereotype.Service;
+
+import javax.servlet.http.HttpServletRequest;
+import java.io.IOException;
+
+import tds.common.ValidationError;
+import tds.student.services.abstractions.IOpportunityService;
+import tds.student.services.data.PageList;
+import tds.student.services.data.TestOpportunity;
+import tds.student.sql.data.OpportunityStatusChange;
+import tds.student.sql.data.OpportunityStatusType;
+import tds.student.sql.repository.remote.ExamRepository;
+import tds.student.web.TestManager;
+
+@Service
+@Scope("prototype")
+public class ReviewTestServiceImpl implements ReviewTestService {
+    private final IOpportunityService opportunityService;
+    private final ExamRepository examRepository;
+    private final ITDSLogger tdsLogger;
+    private final boolean isLegacyCallsEnabled;
+    private final boolean isRemoteCallsEnabled;
+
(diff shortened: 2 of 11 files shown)


student/pom.xml
```diff
@@ -24,7 +24,7 @@
 	<properties>
 		<progman-client.version>3.0.1</progman-client.version>
 		<sb11-mna-client.version>3.0.0</sb11-mna-client.version>
-		<tds-exam-client.version>0.0.14</tds-exam-client.version>
+		<tds-exam-client.version>0.0.15-SNAPSHOT</tds-exam-client.version>
```
jeffjohnson9046, line 27: Will update this to use the release version of the `tds-exam-client` when [the exam service PR for TDS-519](https://github.com/SmarterApp/TDS_ExamService/pull/170) is approved.

student/src/main/java/tds/student/web/handlers/TestShellHandler.java
```diff
@@ -231,31 +236,7 @@ private String parseOutLatencies (Map<String, String> formParams) {
 
     TestOpportunity testOpp = StudentContext.getTestOpportunity ();
 
-    // get responses
```
burzum619, line 234: Why are we removing all this additional logic on the student side? I was under the impression testManager.CheckIfTestComplete() (line 239) and testManager.LoadResponses() (line 236) were already implemented. TestManager.CheckIfTestComplete goes through nearly the same logic flow that the new /review endpoint does (Checking if segments are "satisfied")

student/src/main/java/tds/student/web/handlers/TestShellHandler.java
```diff
@@ -231,31 +236,7 @@ private String parseOutLatencies (Map<String, String> formParams) {
...
-    // check if all visible pages are completed
-    PageList pageList = testManager.GetVisiblePages ();
-
-    if (!pageList.isAllCompleted ()) {
-      String message = "Review Test: Cannot end test because all the groups have not been answered.";
-      _tdsLogger.applicationError (message, "reviewTest", request, null);
-      HttpContext.getCurrentContext ().getResponse ().sendError (HttpStatus.SC_FORBIDDEN, "Cannot end the test.");
-      return null;
-    }
+    ResponseData<String> responseData = reviewTestService.reviewTest(testOpp, request);
```
burzum619: Here we make the call to exam service to check if segments are all satisfied - but are we doing (or do we need to do) the same level of validation above?

student/src/main/java/tds/student/services/ReviewTestServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+        if (isLegacyCallsEnabled) {
+            responseData = legacyReviewTest(testOpportunity, request);
+        }
+
+        if (!isRemoteCallsEnabled) {
+            return responseData;
+        }
+
+        final Optional<ValidationError> maybeValidationError = examRepository.reviewExam(testOpportunity.getOppInstance());
+        if (maybeValidationError.isPresent()) {
```
millerts: To me this block belongs in the Web layer not the service layer.  An error should be sent back from this service and then the controller can do with it what it wants.  In this example it would send the error.

For example, what if we wanted to use this service deep in another service that doesn't have anything to do with HTTP?

student/src/main/java/tds/student/services/ReviewTestServiceImpl.java
```diff
@@ -0,0 +1,128 @@
...
+        // get responses
+        final TestManager testManager = new TestManager(testOpportunity);
+        testManager.LoadResponses(true);
+
+        // check if test length is met
+        testManager.CheckIfTestComplete();
+
+        if (!testManager.IsTestLengthMet()) {
+            // TODO mpatel - Check to see status and message in response and make sure following works
+            final String message = "Review Test: Cannot end test because test length is not met.";
```
millerts: Same comment about this belong in the controller layer

student/src/main/java/tds/student/sql/repository/remote/impl/RemoteExamRepository.java
```diff
@@ -252,6 +255,28 @@ public void approveAccommodations(UUID examId, ApproveAccommodationsRequest appr
...
+    try {
+      restTemplate.put(uriComponentsBuilder.encode().toUri(), null);
+    } catch (HttpClientErrorException hce) {
+      // An unprocessable entity response means that a validation error was returned by the exam service.  Extract the
+      // message and return it as a ValidationError to the caller.
+      if (hce.getStatusCode().equals(HttpStatus.UNPROCESSABLE_ENTITY)) {
+        return Optional.of(objectMapper.readValue(hce.getResponseBodyAsString(), ValidationError.class));
+      }
+    } catch (RestClientException rce) {
+      throw new ReturnStatusException(rce);
```
millerts: @wtritch's been suggesting a small tweak to the handling of the RCE, which I agree with and have been adding.  The below ensures that we can see the reason it failed.

```
     final ReturnStatusException statusException = new ReturnStatusException("Failed to find page with items: " + e.getResponseBodyAsString());
      statusException.getReturnStatus().setHttpStatusCode(500);
      throw statusException;
```

student/src/main/java/tds/student/web/handlers/TestShellHandler.java
millerts: This isn't removed.  @jeffjohnson9046 extracted this logic block to a service since that is probably where this logic should have been in the first place if this application was following a layered architecture..

Replaced here -> https://github.com/SmarterApp/TDS_Student/pull/70/files#diff-06484f1b0d57f1ccd757d75f466f31ffR55

student/src/main/java/tds/student/web/handlers/TestShellHandler.java
millerts: As far as I can tell all the code in the service is carbon copied in the service.  Or at least I didn't see any difference looking in this review.

millerts (changes requested):
I have some design concerns about the service dealing with HTTP level object since to me that seems like a UI/Controller concern.

As to testing the TestManager, in the past whn I tried to test a class that knew how to pull its dependencies I would create a new constructor that passed in its dependencies.  I would then use it in my tests with the hope some day I would be able to remove the magic constructor.  "magic" - when you directly access the spring container for beans.

student/src/main/java/tds/student/web/handlers/TestShellHandler.java
jeffjohnson9046: Right - what @millerts said.  In order to maintain some consistency with how we're handling the switches for legacy/remote calls, the legacy "review test" logic had to be pulled into its own service.

student/src/main/java/tds/student/web/handlers/TestShellHandler.java
jeffjohnson9046: Aside from a little formatting cleanup, this code is identical to what was previously in `TestShellHandler#reviewTest`

student/src/main/java/tds/student/web/handlers/TestShellHandler.java
```diff
@@ -268,12 +249,8 @@ private String parseOutLatencies (Map<String, String> formParams) {
       }
       PerformTestShellAudit (testOpp, testShellAudit, request);
```
burzum619: Do we want to remove this call at line 250?

student/src/main/java/tds/student/web/handlers/TestShellHandler.java
burzum619: The validation I'm referring to in particular is lines 251 - 258 (!pageList.isAllCompleted), where the page groups are iterated over and checked for completion. I would imagine this is in place to prevent the student from moving to the review page without completing the last set of items. Thinking about it a bit further though, this may be slightly redundant. I recall you mentioning that segments are set to "satisfied" when items are being generated in the /updateResponse flow, but that wouldn't necessarily assure that the student actually completed/responded to all items.

jeffjohnson9046: I think we talked about this today, but the validation you're referring to is in the `ReviewTestServiceimpl`: `PageList#getVisiblePages()` is called and then `pageList.isAllCompleted()` is evaluated.

student/src/main/java/tds/student/web/handlers/TestShellHandler.java
millerts: As mentioned offline let's delete this block.

jeffjohnson9046:
Pursuant to this morning's conversation, this PR is being closed and a new one w/refactoring based on our discussion will be made.

PR #77, opened 2017-05-10
Removing unneeded legacy calls in exam submission flow. Renaming Revi…

burzum619:
…ewService to something more generic, since this update call is used by both /review and /submit flow.

A lot of logic has been removed to MasterShellHandler. This is either because it is already handled within ExamService, or because it is not a needed feature (TDS scoring, saving "server latency"). See comments below.

### DIFF
--- a/student/src/main/java/tds/student/services/ExamCompletionServiceImpl.java
+++ b/student/src/main/java/tds/student/services/ExamCompletionServiceImpl.java
@@ -39,8 +39,8 @@ public class ExamCompletionServiceImpl implements ExamCompletionService {
 
         this.opportunityService = opportunityService;
         this.examRepository = examRepository;
-        isLegacyCallsEnabled = false;
-        isRemoteCallsEnabled = true;
+        isLegacyCallsEnabled = legacyCallsEnabled;
+        isRemoteCallsEnabled = remoteExamCallsEnabled;
     }
 
     @Override
--- a/student/src/main/java/tds/student/services/ExamCompletionServiceImpl.java
+++ b/student/src/main/java/tds/student/services/ExamCompletionServiceImpl.java
@@ -50,7 +50,7 @@ public class ExamCompletionServiceImpl implements ExamCompletionService {
         ResponseData<String> responseData = new ResponseData<>(TDSReplyCode.OK.getCode(), "OK", null);
 
         if (isLegacyCallsEnabled) {
-            responseData = legacyReviewTest(testOpportunity, testManager, statusCode);
+            responseData = legacyValidateAndUpdateTest(testOpportunity, testManager, statusCode);
         }
 
         if (!isRemoteCallsEnabled) {
@@ -82,9 +82,13 @@ public class ExamCompletionServiceImpl implements ExamCompletionService {
      * @throws ReturnStatusException In the event of a failure from one of the {@link tds.student.web.TestManager}
      *                               methods
      */
-    private ResponseData<String> legacyReviewTest(final TestOpportunity testOpportunity, final TestManager testManager,
-                                                  final String statusCode)
+    private ResponseData<String> legacyValidateAndUpdateTest(final TestOpportunity testOpportunity, final TestManager testManager,
+                                                             final String statusCode)
         throws ReturnStatusException {
+        if (!ExamStatusCode.STATUS_REVIEW.equals(statusCode) && !ExamStatusCode.STATUS_COMPLETED.equals(statusCode)) {
+            throw new IllegalArgumentException("Can only call the legacy method with a status code of 'review' or 'completed'.");
+        }
+
         // get responses
         testManager.LoadResponses(true);
 
--- a/student/src/main/java/tds/student/services/ExamCompletionServiceImpl.java
+++ b/student/src/main/java/tds/student/services/ExamCompletionServiceImpl.java
@@ -22,34 +22,35 @@ import tds.student.web.TestManager;
 
 @Service
 @Scope("prototype")
-public class ReviewTestServiceImpl implements ReviewTestService {
+public class ExamCompletionServiceImpl implements ExamCompletionService {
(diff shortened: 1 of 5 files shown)


student/src/main/java/tds/student/web/handlers/MasterShellHandler.java
```diff
@@ -654,54 +656,12 @@ public boolean evaluate (Object arg0) {
...
-    if (!pages.isAllCompleted ()) {
-      String message = "Review: A student has tried to complete the test but all the questions are not completed.";
-      _tdsLogger.applicationError (message, "scoreTest", request, null);
-      HttpContext.getCurrentContext ().getResponse ().setStatus (HttpStatus.SC_FORBIDDEN);
-      return new ResponseData<TestSummary> (TDSReplyCode.Denied.getCode (), message, null);
-    }
-
-    // complete test
-    _oppService.setStatus (testOpp.getOppInstance (), new OpportunityStatusChange (OpportunityStatusType.Completed, true));
-
```
burzum619, line 690: Validation done between 667 - 686 is done on Exam side - see the validator JJ added as part of his /reviewTest work.

student/src/main/java/tds/student/web/handlers/MasterShellHandler.java
```diff
@@ -654,54 +656,12 @@ public boolean evaluate (Object arg0) {
...
+    examCompletionService.updateStatusWithValidation(testOpp, new TestManager(testOpp), ExamStatusCode.STATUS_COMPLETED);
+    // Send the exam status to ART
     sendTestStatus (StudentContext.getTestee ().getId (), testOpp.getTestKey (), testOpp.getOppInstance ().getKey (), TestStatusType.COMPLETED);
 
-    // score the test
-    TestScoreStatus scoreStatus = _testScoringService.scoreTest (testOpp.getOppInstance ().getKey (), testOpp.getTestKey ());
```
burzum619, line 694: No need to call `scoreTest` - all this method does (since TDS scoring is not implemented/used) is insert a record into the `qareportqueue` table, which we are no longer using. This same logic is done on the ExamService/ERT side: Exam, when the status is changed to "completed", will submit a message to ERT via that rabbitmq queue/MessageService

student/src/main/java/tds/student/web/handlers/MasterShellHandler.java
```diff
@@ -654,54 +656,12 @@ public boolean evaluate (Object arg0) {
...
-
-    // if we successfully scored the record the server latency
-    if (scoreStatus == TestScoreStatus.Submitted) {
-      ServerLatency latency = ServerLatency.getCurrent (HttpContext.getCurrentContext ());
-      latency.setOperation (ServerLatency.OperationType.Score);
-    }
-
-    // try and get scores
-    return getTestSummary ();
```
burzum619, line 703: No need to call `getTestService()` - its a mess of conditionals that ultimately result in the return status below, regardless of settings. isScoredByTDS is always disabled, and there is nothing to do within this method since scoring is not implemented in TDS.

student/src/main/java/tds/student/web/handlers/MasterShellHandler.java
```diff
@@ -654,54 +656,12 @@ public boolean evaluate (Object arg0) {
...
-      HttpContext.getCurrentContext ().getResponse ().setStatus (HttpStatus.SC_FORBIDDEN);
-      return new ResponseData<TestSummary> (TDSReplyCode.Denied.getCode (), message, null);
-    }
-
-    // complete test
-    _oppService.setStatus (testOpp.getOppInstance (), new OpportunityStatusChange (OpportunityStatusType.Completed, true));
-
+    examCompletionService.updateStatusWithValidation(testOpp, new TestManager(testOpp), ExamStatusCode.STATUS_COMPLETED);
+    // Send the exam status to ART
     sendTestStatus (StudentContext.getTestee ().getId (), testOpp.getTestKey (), testOpp.getOppInstance ().getKey (), TestStatusType.COMPLETED);
```
burzum619: `sendTestStatus` will require some additional attention - I will create a story to cover this work. We need  to send the correct "opportunity/attempt number" when calling this ART REST endpoint.

jeffjohnson9046 (approved):

student/src/main/java/tds/student/services/ExamCompletionServiceImpl.java
```diff
@@ -44,12 +44,13 @@ public ReviewTestServiceImpl(@Qualifier("integrationOpportunityService") final I
...
     @Override
-    public ResponseData<String> reviewTest(final TestOpportunity testOpportunity, final TestManager testManager)
+    public ResponseData<String> updateStatusWithValidation(final TestOpportunity testOpportunity, final TestManager testManager,
+                                                           final String statusCode)
         throws ReturnStatusException {
         ResponseData<String> responseData = new ResponseData<>(TDSReplyCode.OK.getCode(), "OK", null);
 
         if (isLegacyCallsEnabled) {
-            responseData = legacyReviewTest(testOpportunity, testManager);
+            responseData = legacyReviewTest(testOpportunity, testManager, statusCode);
```
millerts: What would happen here if I passed in some other status other than Review?

millerts: Actually, status doesn't matter.  I could pass in Start status and legacy would pass assuming i've answered all the questions.

millerts (approved):

student/src/main/java/tds/student/services/ExamCompletionServiceImpl.java
burzum619: Good point, it would set the simply call the updateStatus endpoint on ExamService, which may or may not be successful, depending on what you pass in. I can add code to validate that the status is either "review" or "completed". Also, renaming this private method

PR #95, opened 2017-07-05
Feature/content service integration

burzum619:
Created RemoteContentRepository to fetch content from the ContentService. Updated existing ContentService to call remote implementation

Since we are no longer supporting legacy, and for the sake of avoiding a ton of duplicate code, I did not create a separate RemoteContentService.


mtrupkin:
obligatory final comments for findItemDocument

mtrupkin (approved):

millerts (approved):

PR #96, opened 2017-07-11
Add idle timeout configuration for the hikari datasource so that conn…

millerts:


student/src/main/resources/performance-context.xml
```diff
@@ -30,6 +30,7 @@
         <property name="maxLifetime" value="1800000" />
         <property name="minimumIdle" value="${performance.datasource.minPoolSize:5}" />
         <property name="maximumPoolSize" value="${performance.datasource.maxPoolSize:20}" />
+        <property name="idleTimeout" value="${performance.datasource.idleTimeout:10000}"/>
```
millerts: This will remove idle connections over time

PR #97, opened 2017-07-15
Feature/s3 resources - Fetching Resources from S3

burzum619:


pom.xml
```diff
@@ -15,7 +15,7 @@
 
     <properties>
         <assertj.version>2.6.0</assertj.version>
-        <item-renderer.version>3.1.1.RELEASE</item-renderer.version>
+        <item-renderer.version>3.1.5-SNAPSHOT</item-renderer.version>
```
burzum619: Will update this to non-snapshot before merge

student/src/main/java/tds/student/web/handlers/DialogFrameHandler.java
```diff
@@ -94,10 +104,10 @@ public String getDialogFrameContent () {
...
     itemRenderGroup.add (new ItemRender (itsDocument, (int) itemKey));
 
-    _pageLayout.setItemRenderGroup (itemRenderGroup);
+    pageLayout.setItemRenderGroup (itemRenderGroup);
 
-    RendererServlet.getRenderedOutput (_pageLayout);
-    System.out.println (_pageLayout.getRenderToString ());
-    return _pageLayout.getRenderToString ();
+    RendererServlet.getRenderedOutput (pageLayout);
+    System.out.println (pageLayout.getRenderToString ());
```
mtrupkin, line 110: writing to standard out

mtrupkin (approved):

millerts (approved):
No issue with the change but I'm not understanding why the resource handler needs to be registered as a servlet rather than a standard Spring handler.  I think it has led to some noise in the item renderer side.

PR #98, opened 2017-07-21
Feature/content service wordlist

mtrupkin:
* add word list support 
* fixes exception when rendering items with audio files caused by browser detection in ITSUrlResolver2.

### DIFF
--- a/student/src/main/java/tds/student/web/handlers/DialogFrameHandler.java
+++ b/student/src/main/java/tds/student/web/handlers/DialogFrameHandler.java
@@ -8,6 +8,7 @@
  ******************************************************************************/
 package tds.student.web.handlers;
 
+import AIR.Common.Web.BrowserParser;
 import AIR.Common.Web.Session.Server;
 import AIR.Common.Web.WebHelper;
 import org.apache.commons.lang3.StringUtils;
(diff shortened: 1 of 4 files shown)


student/src/main/java/tds/student/web/handlers/DialogFrameHandler.java
```diff
@@ -91,7 +92,7 @@ public String getDialogFrameContent () {
     {
       final AccLookup languageAccommodation = new AccLookup();
       languageAccommodation.add ("Language", contentLanguage);
-      itsDocument = contentRepository.findItemDocument(studentHelpFile, languageAccommodation, Server.getContextPath());
+      itsDocument = contentRepository.findItemDocument(studentHelpFile, languageAccommodation, Server.getContextPath(), new BrowserParser().isSupportsOggAudio());
```
millerts: I think this handler should be using the `ContentService` and not the repository.  I know you're carrying it forward, but I personally prefer handlers to only talk to services especially since transaction boundaries are added on service methods.

millerts (approved):
LGTM outside of that small change.

Obligatory comment about making sure to update snapshots as things are released.

PR #101, opened 2017-08-10
Feature/spring upgrade

mtrupkin:


student/pom.xml
```diff
@@ -259,14 +257,12 @@
...
 			<artifactId>spring-jdbc</artifactId>
-			<version>${spring.version}</version>
 		</dependency>
 
 		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-test</artifactId>
-			<version>${spring.version}</version>
-			<scope>test</scope>
+			<scope>compile</scope>
```
millerts, line 266: Can you add a note in the XML that explains why this isn't a test scope?  Something like `Classes in spring-test are currently used within non test code`

student/src/main/java/tds/student/web/handlers/TestResponseHandler.java
```diff
@@ -49,6 +50,7 @@
 
 @Controller
 @Scope ("request")
+@RequestMapping(produces = MediaType.APPLICATION_JSON_VALUE)
```
millerts, line 53: I'm a bit concerned with the class level request mapping forcing JSON here.  For example, https://github.com/SmarterApp/TDS_Student/pull/101/files#diff-ca388e072dbdc1dca4e729a06468d52aR110 writes XML to the response.  Does that produce json now?

millerts: I understand that this fixes potential other methods in this class.  Just that it seems like everything doesn't actually return Json.

student/src/main/java/tds/student/services/ItemScoringService.java
```diff
@@ -72,6 +75,7 @@
 
 @Service("legacyItemScoringService")
 @Scope ("prototype")
+@RequestMapping(produces = MediaType.APPLICATION_JSON_VALUE)
```
millerts: Is this still being called in the UI?

millerts (commented):
I understand the changes but I wish we knew why we needed to force everything to return JSON in this way now and it worked prior to the upgrade

PR #109, opened 2017-08-30
TDS-1189: Remove hard-coding width from additional locations

jeffjohnson9046:
[TDS-1189](https://jira.fairwaytech.com/browse/TDS-1189):  The width of the illustration glossary was also enforced (with an `!important`) at each zoom level.  Also noticed that sometimes the verbiage would extend past the boundaries of the modal window.  Took the following action:

* Removed the `width: [size in pixels] !important;` from each zoom-level class
* Added `overflow: hidden;` to prevent the "header" text directly above the image from extending beyond the width of the modal


jeffjohnson9046:
This PR is out-of-date and will be addressed by another PR.

